PR #12, opened 2024-05-14
setup workflow to run rust test

Iamdavidonuh:


ahnlich/server/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
+#![allow(clippy::size_of_ref)]
```
Iamdavidonuh, line 2: @deven96 allowing the size of ref for now. We would need to come up with the right implementation for calculating the size of the Store

deven96: Yeah... all the sizing already feels very off

deven96 (approved):

PR #16, opened 2024-05-18
SImple Server Setup

### LINKED ISSUE #15: Basic server entrypoint
This ticket should aim to do a few things. The first being 
- Introducing a basic tokio main to the server that binds to a TCP endpoint and probably just echoes whatever is sent
- Create clap configuration for the server (can add clap to the entire workspace Cargo.toml and reuse in server as we'd likely have clap also down the line for ahnlich-ai). This configuration should accept a few things with some defaults set such as
   - host: Host to bind to
   - port: Port to bind to (we can default to 1369)
   - allow-persistence: allows to occasionally persist stores to disk
   - persistence-location: path to save persistence file to. Default to something reasonable like using the same directory where the binary is. Also only use if allow-persistence is given
   - peristence-interval-ms: persistence interval given in milliseconds. This would control how often we try to persist (For starters we can default to something reasonable like every 5 minutes). For implementation every store would have to figure out if any update to the store internals has occurred since last update and trigger the persistence then cc @lordsarcastic ... also depends on allow-persistence
 There are other command line flags over time that we'd add but let's start with these

Iamdavidonuh:
closes #15


ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
+use std::error::Error;
 
-#[cfg(test)]
-mod tests;
+use clap::{ArgAction, Args, Parser, Subcommand};
+use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use tokio::net::{TcpListener, TcpStream};
+
+async fn run_server() -> std::io::Result<()> {
+    let listener = TcpListener::bind("127.0.0.1:1369").await?;
```
deven96, line 15: Make use of the port and host here

ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
 
-fn main() {
-    println!("Hello, world!");
+    loop {
+        let (stream, connect_addr) = listener.accept().await?;
+        println!("Connecting to {}", connect_addr);
```
deven96, line 19: Most likely introduce log in the workspace instead of using any print or println so the log level can be easily toggled

ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
+        });
+    }
 }
+
+async fn process_stream(stream: TcpStream) -> Result<(), tokio::io::Error> {
+    stream.readable().await?;
+    let mut reader = BufReader::new(stream);
+    loop {
+        let mut message = String::new();
+        let _ = reader.read_line(&mut message).await?;
```
deven96, line 33: Nice.... this would be where we subsequently then deserialize the message coming in (as bytes) using bincode

ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
+        reader.get_mut().write_all(message.as_bytes()).await?;
+        message.clear();
+    }
+}
+
+#[derive(Parser)]
```
deven96, line 39: Move all the clap command line stuff to its own mod

ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
+    #[arg(long, requires_if("true", "enable_persistence"))]
+    persist_location: Option<std::path::PathBuf>,
+
+    /// persistence intervals in milliseconds
+    #[arg(long, default_value_t = 1000 * 60 * 5)]
+    persistence_intervals: u64,
+}
+
+#[derive(Args, Debug)]
+struct Persistence {}
```
deven96, line 68: What's this?

deven96 (changes requested):

ahnlich/server/src/main.rs
```diff
@@ -5,10 +5,85 @@ mod engine;
...
 }
+
+async fn process_stream(stream: TcpStream) -> Result<(), tokio::io::Error> {
+    stream.readable().await?;
+    let mut reader = BufReader::new(stream);
+    loop {
+        let mut message = String::new();
+        let _ = reader.read_line(&mut message).await?;
+        reader.get_mut().write_all(message.as_bytes()).await?;
+        message.clear();
```
deven96, line 35: Add a basic test for your echo case... obviously we'd soon override it and it would start failing but that would push us to add more appropriate tests when it does

ahnlich/server/src/main.rs
Iamdavidonuh: That was the plan before you marked it as ready for review 👀

deven96 (approved):

PR #27, opened 2024-06-08
Generate type definitions for Client Libraries

### LINKED ISSUE #17: Generate type definitions for Client Libraries
We already have defined types that are potentially going to be shared across the client and server libraries in the types crate.

This issue will try to create underlying type definitions to be used by client libraries.

[Serde Reflection](https://docs.rs/serde-reflection/latest/serde_reflection/)
[Serde Generate](https://docs.rs/serde-generate/latest/serde_generate/)

### Acceptance Criteria
Test that the client(mini implementation) and server are both able to communicate using the types generated.
i:e: generated types in a language are send and properly interpreted by the rust server.

Iamdavidonuh:
This PR Addresses some points in #17


ahnlich/types/src/query.rs
```diff
@@ -69,6 +76,25 @@ pub enum Query {
...
+    S: serde::Serializer,
+{
+    serializer.serialize_u64(value.get() as u64)
+}
+
+fn deserialize_non_zero_usize<'de, D>(deserializer: D) -> Result<NonZeroUsize, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let mut value = u64::deserialize(deserializer)?;
```
deven96, line 90: This should be usize not u64

ahnlich/types/src/query.rs
```diff
@@ -69,6 +76,25 @@ pub enum Query {
...
+    if value < 1 {
+        value += 1
```
deven96, line 92: We shouldn't sneakily add stuff for them
I'd rather the deserialization fails than do something sneaky like turning 0s to 1s

deven96: We might later catch deserialization errors to report back to the client (although we don't currently do that, we just break)

ahnlich/types/src/query.rs
```diff
@@ -69,6 +76,25 @@ pub enum Query {
     ListClients,
     Ping,
 }
+fn serialize_non_zero_usize<S>(value: &NonZeroUsize, serializer: S) -> Result<S::Ok, S::Error>
```
deven96, line 79: They don't have serialise for usize? Not u64?

Iamdavidonuh: @deven96  No,serde doesn't impl serialize for usize which makes sense because usize is based on system arch(32 or 64)

ahnlich/types/src/query.rs
Iamdavidonuh: The issue here is, tracer adds a default value of 0 abandoning the default value specified causing nonzerousize to break

Iamdavidonuh: Let me try creating a sample, just to confirm

ahnlich/types/src/query.rs
deven96: That makes sense... resolved

ahnlich/types/src/query.rs
Iamdavidonuh: Yeah, creating samples resolves this, so no need for extra serialization and deserialization functions

ahnlich/typegen/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+        store: sample_store_name.clone(),
+        condition: test_predicate_condition.clone(),
+    };
+    let deletepred_variant = Query::DelPred {
+        store: sample_store_name.clone(),
+        condition: test_predicate_condition.clone(),
+    };
+
+    let _ = tracer.trace_value(&mut samples, &create_store).unwrap();
+    let _ = tracer.trace_value(&mut samples, &create_store).unwrap();
```
deven96, line 83: Why trace create_store twice?

deven96 (approved):

ahnlich/typegen/src/tracers/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+            Language::Golang => {
+                let query_file =
+                    std::fs::File::create(self.output_dir.join(format!("{}.go",file_name)))
+                    .expect("Failed to create file");
+                let mut buffer: BufWriter<File> = std::io::BufWriter::new(query_file);
+                serde_generate::golang::CodeGenerator::new(&config)
+                    .output(&mut buffer, registry)
+            }
+            Language::Ocaml => {
+                let query_file =
```
deven96, line 102: You can do this entire thing outside

If there is a  

```rust
struct OutputFile {
 output_file: std::path::PathBuf,
 language: Language,
}

impl OutputFile {
 fn generate(&self, config, registry) -> std::fs::File {
  let query_file = std::fs::File::create(&self.output_file).expect("Failed to create typegen output file");
  let mut buffer = BufWriter<File> = std::io::BufWriter::new(query_file);
  match self.language {
  Language::Python => serde_generate::python3::CodeGenerator::new(config).output(&mut buffer, registry),
 Language::Golang => serde_generate::go::CodeGenerator::new(config).output(&mut buffer, registry), 
 }
}

impl From<(&std::path::PathBuf, &Language)> for OutputFile {
 fn from((file_dir, file_name, language): (&std::path::PathBuf, &str, &Language)) -> OutputFile {
  let extension = match language {
  Language::Python => "py",
  Language::Go => "go",
  ...
  };
  OutputFile(file_dir.join(format!("{file_name}.go")))
 }
}
```

And then process typegen can look a bit simpler like

```rust
fn process_type_gen(file_dir, file_name) {
   if !file_dir.is_dir() {
   // maybe this would have already been checked at a higher layer
   // else check here and panic
  }
  let query_file: OutputFile = (file_dir, file_name, &self.language).into();
 query_file.generate(&config, &registry);
}
// or modify generate_type_def_for_language to do this directly
```

deven96 (changes requested):

Iamdavidonuh:
@deven96  This change will break the typesgen test since a change was made in the types crate.

[Diff](https://github.com/deven96/ahnlich/pull/27/commits/b272e19bbc0b6d3b19b397de62a7e6dddb034e5b)

deven96 (approved):

PR #32, opened 2024-06-12
Adding AI musings and In/NotIn ops

deven96:
A bit repetitive to have the negate symbols literally just do the opposites of what the normal symbols are doing but in a verbose manner but I rather the clear splits for now anyways

Reducing time spent turning a `StoreKey` into `StoreKeyId` by using [blake3](https://github.com/BLAKE3-team/BLAKE3?tab=readme-ov-file)


ahnlich/db/src/engine/predicate.rs
```diff
@@ -255,25 +255,32 @@ impl PredicateIndex {
...
-        predicate_op: &'a PredicateOp,
-        value: &'a MetadataValue,
-    ) -> StdHashSet<StoreKeyId> {
+    fn matches(&self, predicate: &Predicate) -> StdHashSet<StoreKeyId> {
         let pinned = self.0.pin();
-        match predicate_op {
-            PredicateOp::Equals => {
+
+        match predicate {
+            Predicate::Equals { value, .. } => {
```
Iamdavidonuh, line 262: Probably a Nit Pick or a question, but isn't this suppose to be the key?

ahnlich/db/src/engine/predicate.rs
```diff
@@ -255,25 +255,32 @@ impl PredicateIndex {
...
                 if let Some(set) = pinned.get(value) {
                     set.pin().iter().cloned().collect::<StdHashSet<_>>()
                 } else {
                     StdHashSet::new()
                 }
             }
-            PredicateOp::NotEquals => pinned
+            Predicate::NotEquals { value, .. } => pinned
```
Iamdavidonuh, line 269: dito

ahnlich/db/src/engine/predicate.rs
deven96: Nah its the value you compare against

Iamdavidonuh (approved):

Issue #9, opened 2024-05-12
Thoughts and standard spec on Persistence


deven96:
Expanding on some of the things outlined in [draft](https://github.com/deven96/ahnlich/blob/main/docs/persistence.md)

PR #33, opened 2024-06-14
Basic persistence capabilities with separate thread

deven96:
Creates a v1 for #9

But as a wise man once said.... there's nothing more permanent than a temporary solution

### DIFF
--- a/ahnlich/db/src/persistence/mod.rs
+++ b/ahnlich/db/src/persistence/mod.rs
@@ -1,9 +1,10 @@
 use crate::engine::store::Stores;
 use std::fs::File;
-use std::fs::OpenOptions;
+use std::path::Path;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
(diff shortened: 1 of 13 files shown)


ahnlich/db/src/persistence/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[tracing::instrument(skip(self))]
+    pub(crate) async fn monitor(&mut self, shutdown_guard: ShutdownGuard) {
+        loop {
+            select! {
+                _  = shutdown_guard.cancelled() => {
+                    break;
+                }
+                has_potential_write = self.has_potential_write() => {
+                    if has_potential_write {
+                        let writer = if let Ok(file) = OpenOptions::new().create(true).truncate(true).write(true).open(&self.persist_location) {
```
deven96, line 50: I don't like this truncation as I stare at it again because we could loose all data in a crash scenario

Marking this to persist to a separate temp location and then replace the original file after

Iamdavidonuh (approved):

Issue #31, opened 2024-06-12
Python Client Set up and Automation 


deven96:
- [x] Ensure that `server_response.Version` always matches with the `pyproject.toml` , keeping in mind that the database always communicates with only the exact version
- [x] Introduce automatic version bump that goes through rust workspace Cargo.toml, pyproject.toml, and then goes on to automatically package and deploy the python client using twine to `pypi`. We can start out by deploying to `test.pypi.org` until we are actually ready to deploy all clients to their final environments but we should already have the pipelines up and running
- [x] Write multiple integrations tests that include calls to an actual database service, creating entries in the database, deleting, searching, indexing, e.t.c
- [x] Given that we are going to introduce an `ai` layer, we want the client to cater for both, so for now, requests should be separated into `ai` and `db` modules. This is because later on we will be generating some `ai` specific types and bindings
- [x] To establish TCP connections, user should be able to able to define the host and port to connect. We should then be able to provide them pooling abstractions (perhaps use any existing library like [generic-connection-pool](https://pypi.org/project/generic-connection-pool/)). We should also be able to allow pipelining i.e where we send multiple requests at once using the `query.ServerQuery` to the selected tcp connection from the pool or allow sending a single request to any tcp connection in the pool
- [x] Add a proper README to the Python client and make sure the methods are well documented. We can document everything for now on the README to prevent having to release a separate python documentation

Related to #17

PR #37, opened 2024-06-24
Create Connection pool for Ahnlich DB client

Iamdavidonuh:
Relates to #31


sdk/ahnlich-client-py/ahnlich_client_py/config.py
```diff
@@ -1,8 +1,34 @@
...
+
+    def __init__(self) -> None:
+
+        self.HEADER = b"AHNLICH;"
+        self.BUFFER_SIZE = 1024
+        self.PACKAGE_NAME = "ahnlich-client-py"
+        self.BASE_DIR = Path(__file__).resolve().parent.parent
+        self.AHNLICH_BIN_DIR = self.BASE_DIR.parent.parent / "ahnlich"
+        # connection pool
+        self.POOL_IDLE_TIMEOUT = float(os.environ.get("AHNLICH_IDLE_TIMEOUT", 30.0))
```
deven96, line 15: we don't need to set these as environmental variables to be honest. See below

sdk/ahnlich-client-py/ahnlich_client_py/protocol.py
```diff
@@ -1,31 +1,37 @@
...
 from ahnlich_client_py.internals import query, server_response
 
 
 class AhnlichProtocol:
     def __init__(self, address: str, port: int, timeout_sec: float = 5.0):
-        self.address = address
+        self.address = IPv4Address(address)
         self.port = port
-        self.conn = self.connect()
+        self.connection_pool = self.create_connection_pool()
```
deven96, line 16: Instead of using the environmental variables, allow the connection details such as `max_lifetime`, `total_max_size`, e.t.c to be Option[T] and passed into the init and flow into `self.create_connection_pool`, then within create_connection_pool, unwrap them and assign a default, this lets us do something like

```py
client = AhnlichDBClient(
    host="...",
    pool=AhnlichDBPoolSettings(
        max_lifetime_secs=30|None,
        total_max_size=10|None,
     ) | None,
)
```

I think python dataclass can give you that ability to specify defaults and then have people override them

sdk/ahnlich-client-py/ahnlich_client_py/libs.py
```diff
@@ -11,3 +11,17 @@ def create_store_key(data: typing.List[float], v: int = 1) -> query.Array:
     dimensions = (st.uint64(np_array.shape[0]),)
     store_key = query.Array(v=st.uint8(v), dim=dimensions, data=np_array.tolist())
     return store_key
+
```
deven96, line 14: Unused right?

deven96 (changes requested):

sdk/ahnlich-client-py/ahnlich_client_py/protocol.py
Iamdavidonuh: Done

sdk/ahnlich-client-py/ahnlich_client_py/libs.py
Iamdavidonuh: removed

deven96 (approved):

PR #43, opened 2024-06-26
Rust Client Implementation for DB

### LINKED ISSUE #36: Rust Client Setup
- [x] Should include an ai and db module as client can be used to talk to the ai proxy or talk to the db directly
- [x] Might include integrations tests for the db and ai module
- [x] Given that we are going to introduce an `ai` layer, we want the client to cater for both, so for now, requests should be separated into `ai` and `db` modules. This is because later on we will be generating some `ai` specific types and bindings
- [x] To establish TCP connections, user should be able to able to define the host and port to connect. We should then be able to provide them pooling abstractions (perhaps use any existing library like [r2d2](https://docs.rs/r2d2/latest/r2d2/)). We would abstract away pipelining i.e where we send multiple requests at once using the `query.ServerQuery` to the selected tcp connection from the pool or allow sending a single request to any tcp connection in the pool. A pipeline has a reference to the original client as it needs it to be able to execute
- [x] Add a proper README to the Rust client and make sure the methods are well documented with docstrings for rustdocs.
- [x] Badge on main README showing Rust client version
- [ ] Test deploy to crates.io with a dry run and set up pipeline to actually deploy whenever ready

deven96:
To resolve #36

### DIFF
--- a/.github/workflows/rust_tag_and_deploy.yml
+++ b/.github/workflows/rust_tag_and_deploy.yml
@@ -1,0 +1,75 @@
+name: Ahnlich Python Client Tag and Deploy
+
+on:
+  pull_request:
+    branches: ["main"]
+    types:
+      - closed
+
+jobs:
+  check_version_changes_and_tag:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    outputs:
+      client_version: ${{ steps.get_version.outputs.CLIENT_VERSION }}
+      tag_version: ${{ steps.get_version.outputs.TAG_VERSION }}
+    steps:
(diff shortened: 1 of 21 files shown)


.github/workflows/rust_tag_and_deploy.yml
```diff
@@ -0,0 +1,75 @@
+name: Ahnlich Python Client Tag and Deploy
```
Iamdavidonuh, line 1: Rust client

Iamdavidonuh (changes requested):

Iamdavidonuh (approved):

PR #50, opened 2024-06-29
WIP: Adding criterion testing and pipeline

deven96:


deven96:
Breaks right now as the database benchmark is not yet on main

PR #71, opened 2024-07-21
Add AI Proxy client 

Iamdavidonuh:
- Add ahnlich ai proxy client tests
- Isolate ahnlich db client tests


sdk/ahnlich-client-py/README.md
```diff
@@ -315,9 +330,185 @@ response = client.delete_predicate(
...
+    keys=[key]
+)
+```
+
+
+
 
 ## Bulk Requests
-The client has the ability to send multiple requests at once, and these requests will be handled sequentially. The builder class takes care of this. The response is a list of all individual request responses.
+The clients has the ability to send multiple requests at once, and these requests will be handled sequentially. The builder class takes care of this. The response is a list of all individual request responses.
```
deven96, line 511: *Clients have the ability

deven96 (approved):

PR #77, opened 2024-07-31
AIProxy Set Command Preprocessing actions for storeinputs

Iamdavidonuh:


ahnlich/ai/src/engine/ai/mod.rs
```diff
@@ -0,0 +1,10 @@
+mod models;
+use ahnlich_types::keyval::{StoreInput, StoreKey};
+use models::ModelInfo;
+use std::num::NonZeroUsize;
+
+pub trait AIModelManager {
+    fn embedding_size(&self) -> NonZeroUsize;
```
deven96, line 7: Can be a default implementation if it's already coming from the ModelInfo or get rid of it entirely even

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -0,0 +1,42 @@
...
+    keyval::{StoreInput, StoreKey},
+};
+use ndarray::Array1;
+use serde::{Deserialize, Serialize};
+use std::num::NonZeroUsize;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ModelInfo {
+    pub name: String,
+    pub embedding_size: usize,
```
deven96, line 13: Can make a NonZeroUsize and be done with

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn model_ndarray(&self, storeinput: &StoreInput) -> StoreKey {
+        let length = storeinput.len() as f32;
+        StoreKey(Array1::from_iter(0..self.model_info().embedding_size).mapv(|v| v as f32 * length))
+    }
+
+    fn model_info(&self) -> ModelInfo {
+        match self {
+            AIModel::Llama3 => ModelInfo {
+                name: String::from("Llama3"),
+                embedding_size: 100,
```
deven96, line 32: `nonzero!()` macro will get you compile time failure here if a zero is passed in so it should be fine to use

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+use ndarray::Array1;
+use serde::{Deserialize, Serialize};
+use std::num::NonZeroUsize;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ModelInfo {
+    pub name: String,
+    pub embedding_size: usize,
+    pub input_type: AIStoreInputTypes,
```
deven96, line 14: Singular InputType

ahnlich/ai/src/engine/store.rs
```diff
@@ -57,14 +58,18 @@ impl AIStoreHandler {
...
+        index_model: AIModel,
     ) -> Result<(), AIProxyError> {
         if self
             .stores
             .try_insert(
                 store_name.clone(),
-                Arc::new(AIStore::create(store_name.clone(), model.clone())),
+                Arc::new(AIStore::create(
+                    store_name.clone(),
+                    query_model.clone(),
```
deven96, line 70: No need to clone the models, we are not reusing below

ahnlich/ai/src/engine/store.rs
```diff
@@ -100,28 +106,57 @@ impl AIStoreHandler {
...
     ) -> Result<(StoreKey, StoreValue), AIProxyError> {
         let metadata_key = &*AHNLICH_AI_RESERVED_META_KEY;
         if store_value.contains_key(metadata_key) {
             return Err(AIProxyError::ReservedError(metadata_key.to_string()));
         }
         let store = self.get(store_name)?;
 
-        let store_key = store.model.model_ndarray(&store_input);
-        let metadata_value: MetadataValue = store_input.into();
+        let store_input_type: AIStoreInputTypes = store_input.clone().into();
```
deven96, line 124: In between these two is where we would ensure that we even started the proxy with that particular `AIModel` support

deven96, line 124: No need to clone the entire store input just to turn it into an `AIStoreInputType` , especially because these inputs can be large (i.e strings and image buffers), you can instead have a `From<&StoreInput> for AIStoreInputType`

ahnlich/ai/src/engine/store.rs
```diff
@@ -100,28 +106,57 @@ impl AIStoreHandler {
...
 
     /// Converts storeinput into a tuple of storekey and storevalue.
-    /// Fails if the type of storeinput does not match the store type
+    /// Fails if the store input type does not match the store index_type
     #[tracing::instrument(skip(self))]
     pub(crate) fn store_input_to_store_key_val(
         &self,
         store_name: &StoreName,
         store_input: StoreInput,
         store_value: &StoreValue,
```
deven96, line 115: Why are we passing a ref here rather than owned?

ahnlich/ai/src/engine/store.rs
```diff
@@ -100,28 +106,57 @@ impl AIStoreHandler {
...
+    #[tracing::instrument(skip(self))]
+    pub(crate) fn create_store_key(
+        &self,
+        store_input: StoreInput,
+        index_store_model: &AIModel,
+        preprocess_action: &PreprocessAction,
+    ) -> Result<StoreKey, AIProxyError> {
+        let mut store_input = store_input;
+
+        self.process_store_input(preprocess_action, &mut store_input, index_store_model)?;
```
deven96, line 155: Put a comment on this ... I personally find these implicit mutations of variables just to use them immediately after a bit icky... however that's a nit and this isn't the final form especially because what the preprocess actions will return is actually some new enum containing processed string tokens or resized image content

ahnlich/ai/src/engine/store.rs
```diff
@@ -185,20 +220,95 @@ impl AIStoreHandler {
...
     }
+
+    #[tracing::instrument(skip(self))]
+    pub(crate) fn process_store_input(
+        &self,
+        process_action: &PreprocessAction,
+        input: &mut StoreInput,
+        index_model: &AIModel,
+    ) -> Result<(), AIProxyError> {
+        match process_action {
```
deven96, line 231: I think a match over (process_action, store_input) would be cleaner and you can error in the right places e.g

```rust
match (process_action, store_input) {
    (PreprocessAction::Image(_), StoreInput::RawString(_)) => Err("Used Image preprocessing for String type"),
    ...
}
```

ahnlich/ai/src/error.rs
```diff
@@ -7,12 +7,38 @@ pub enum AIProxyError {
...
+    #[error("Cannot Query Using Input. Store expects [{store_query_model_type}], but input type [{storeinput_type}] was provided")]
+    StoreQueryTypeMismatchError {
+        store_query_model_type: AIStoreInputTypes,
+        storeinput_type: AIStoreInputTypes,
+    },
+    #[error("Cannot Set Input. Store expects [{store_index_model_type}], input type [{storeinput_type}] was provided")]
+    StoreSetTypeMismatchError {
+        store_index_model_type: AIStoreInputTypes,
+        storeinput_type: AIStoreInputTypes,
+    },
```
deven96, line 29: There should be a PreprocessMismatchError as well

ahnlich/ai/src/tests/aiproxy_test.rs
```diff
@@ -606,7 +627,8 @@ async fn test_ai_proxy_binary_store_actions() {
     let message = AIServerQuery::from_queries(&[
         AIQuery::CreateStore {
             store: store_name.clone(),
-            model: AIModel::Llama3,
+            query_model: AIModel::Llama3,
+            index_model: AIModel::DALLE3,
```
deven96, line 631: The query model and index model have a mismatch in embedding size, why is the createstore still having an Ok

ahnlich/types/src/ai/mod.rs
```diff
@@ -1,29 +1,41 @@
...
-    // TODO: Consider making this an option of nonzerousize
-    pub fn embedding_size(&self) -> NonZeroUsize {
-        NonZeroUsize::new(3).expect("Failed to get embedding size")
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum AIStoreInputTypes {
+    RawString,
+    Image,
+}
+
+impl From<StoreInput> for AIStoreInputTypes {
```
deven96, line 25: &StoreInput since the content is never really used

deven96 (changes requested):

ahnlich/ai/src/engine/store.rs
deven96:
```rust
// allowed_input is a method that effectively tells us either the max amount of string tokens this model
// accepts as inputs or the image size it accepts as input
let preprocessed_input = self.preprocess_input(action, store_input, index_store_model.allowed_input())?;
let store_key = index_store_model.model_ndarray(preprocessed_input);
```

ahnlich/ai/src/error.rs
```diff
@@ -7,12 +7,38 @@ pub enum AIProxyError {
...
+
+    #[error("Max Token Exceeded. Model Expects [{model_embedding_size}], input type was [{input_token_size}] ")]
+    TokenExceededError {
+        model_embedding_size: usize,
```
deven96, line 33: Model embedding size has nothing to do with tokens exceeded

Model embeddings is the output of the model, while preprocessing is basically us trying to get the user input to fit the allowed input of the model

deven96 (approved):

PR #80, opened 2024-08-14
Create Store Validation

Iamdavidonuh:
- On creating a store, validate index and query models against supported models selected on AIProxy startup. We error if either the index or query models don't exist in the supported model's Vec.
- Added a new field max_token to specify the maximum token a model would accept
- Swap out AIModelManager Trait for a Model Enum of Image and Text variants, subsequent variants can be added when we support other models
- Added Subcommand to display supported models in verbose or limited mode

### DIFF
--- a/ahnlich/ai/src/engine/ai/mod.rs
+++ b/ahnlich/ai/src/engine/ai/mod.rs
@@ -1,1 +1,7 @@
 pub mod models;
+use once_cell::sync::Lazy;
+
+use ahnlich_types::ai::AIModel;
+
+pub(crate) static AHNLICH_AI_SUPPORTED_MODELS: Lazy<Vec<AIModel>> =
+    Lazy::new(|| vec![AIModel::Llama3, AIModel::DALLE3]);
--- a/ahnlich/ai/src/engine/ai/mod.rs
+++ b/ahnlich/ai/src/engine/ai/mod.rs
@@ -1,5 +1,5 @@
-mod models;
 use ahnlich_types::keyval::{StoreInput, StoreKey};
+pub(crate) mod models;
 use models::ModelInfo;
 use std::num::NonZeroUsize;
 
--- a/ahnlich/ai/src/engine/ai/mod.rs
+++ b/ahnlich/ai/src/engine/ai/mod.rs
@@ -1,10 +1,1 @@
-use ahnlich_types::keyval::{StoreInput, StoreKey};
-pub(crate) mod models;
-use models::ModelInfo;
-use std::num::NonZeroUsize;
-
-pub trait AIModelManager {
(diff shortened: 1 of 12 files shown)


ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -1,46 +1,72 @@
...
-        StoreKey(
-            Array1::from_iter(0..self.model_info().embedding_size.into())
-                .mapv(|v| v as f32 * length),
-        )
+        StoreKey(Array1::from_iter(0..self.embedding_size().into()).mapv(|v| v as f32 * length))
     }
 
-    fn model_info(&self) -> ModelInfo {
+    /// Returns either the max token size(Text) or max image dimension size(Image) of a model
+    pub fn max_accepted_size(&self) -> NonZeroUsize {
```
deven96, line 61: Don't use one function for this

ahnlich/ai/src/engine/store.rs
```diff
@@ -61,6 +65,11 @@ impl AIStoreHandler {
         query_model: AIModel,
         index_model: AIModel,
     ) -> Result<(), AIProxyError> {
+        if !self.supported_models.contains(&(&query_model).into())
+            || !self.supported_models.contains(&(&index_model).into())
+        {
+            return Err(AIProxyError::AIModelNotInitialized);
```
deven96, line 71: You also need to check that their embedding sizes are the same

ahnlich/ai/src/engine/ai/mod.rs
```diff
@@ -1,10 +1,7 @@
...
+use once_cell::sync::Lazy;
 
-pub trait AIModelManager {
-    fn embedding_size(&self) -> NonZeroUsize;
-    fn model_ndarray(&self, storeinput: &StoreInput) -> StoreKey;
-    fn model_info(&self) -> ModelInfo;
-}
+use ahnlich_types::ai::AIModel;
+
+pub(crate) static AHNLICH_AI_SUPPORTED_MODELS: Lazy<Vec<AIModel>> =
```
deven96, line 6: Supported models are all the models inside AIModel enum... so rather than do this, look into strum to be able to iterate an enum

Iamdavidonuh:
@deven96, Here's how the output looks like

Screenshot 2024-08-18 at 9 57 36 PM
Screenshot 2024-08-18 at 9 58 01 PM

deven96:
> @deven96, Here's how the output looks like
> 
> Screenshot 2024-08-18 at 9 57 36 PM Screenshot 2024-08-18 at 9 58 01 PM

Awesome

deven96 (approved):

PR #90, opened 2024-08-28
Prepare Ahnlich Binary Release

### LINKED ISSUE #89: AI Proxy and DB Version
We want to be able to deploy ahnlich aiproxy and db when their versions change. The versions are different from the protocol version and should reflect as such.

- [x] Have a subcommand to display version of aiproxy and db
- [x] Bumping protocol version also bumps up aiproxy and db version
- [x] Prepare workflow for  releasing binaries

Iamdavidonuh:
Closes #89

### DIFF
--- a/ahnlich/Cargo.toml
+++ b/ahnlich/Cargo.toml
@@ -7,7 +7,7 @@ members = [
   "typegen",
   "utils",
   "similarity",
-  "ai",
+  "ai", "task-manager",
 ]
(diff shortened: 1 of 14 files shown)


deven96 (changes requested):
Once we've factored in Darwin and windows, we can call it a day

deven96 (dismissed):

deven96 (approved):

PR #92, opened 2024-08-31
Setup AI Models

HAKSOAT:
This PR has made the following key changes thus far:

- Added download support for 4 text models and 2 image models
- Added a `--model-cache-location` parameter for the cli
- Partially integrated the use of the FastEmbed package
- Caching for models, to avoid multiple downloads, this takes advantage of Huggingface's caching algorithm.

How to test:

```cargo run --bin ahnlich-ai run --supported-models all-minilm-l6-v2```

The above will download to the default location of `~/.ahnlich/models`

```cargo run --bin ahnlich-ai run --supported-models all-minilm-l6-v2 --model-cache-location ~/anywhere```

The above will download to the `~/anywhere` location.

### CODE REVIEW

File: ahnlich/ai/src/engine/ai/models.rs [hunk 1]
```diff
@@ -31,12 +31,50 @@ impl From<&AIModel> for Model {
...
+            AIModel::BGEBaseEnV15 => Self::Text {
+                name: String::from("BGEBase-En-v1.5"),
+                description: String::from("BAAI General Embedding model with English support, base scale, version 1.5."),
+                embedding_size: nonzero!(768usize),
+                // Token size source: https://huggingface.co/BAAI/bge-large-en/discussions/11#64e44de1623074ac850aa1ae
+                max_input_tokens: nonzero!(512usize),
+            },
+            AIModel::BGELargeEnV15 => Self::Text {
+                name: String::from("BGELarge-En-v1.5"),
+                description: String::from("BAAI General Embedding model with English support, large scale, version 1.5."),
```
deven96, line 62: Do we include model hash in the info so we can confirm that the downloaded model is valid?

ahnlich/ai/src/engine/ai/providers/providers.rs
```diff
@@ -0,0 +1,25 @@
+use std::path::PathBuf;
+use ahnlich_types::ai::AIModel;
+use crate::engine::ai::providers::fastembed::FastEmbedProvider;
+
+
+pub enum ModelProviders {
```
deven96, line 6: Perhaps this is a binary argument i.e the model provider you want to use for download is `fast-embed`.... we can also set the default to fast-embed.... else if it is actually on a model-by-model use-case then perhaps we also add it to the `Model` enum within `Self::Text` and `Self::Image`?

ahnlich/ai/src/engine/ai/providers/fastembed.rs
```diff
@@ -0,0 +1,71 @@
...
+
+
+impl FastEmbedProvider {
+    pub fn new(cache_location: PathBuf) -> Self {
+        FastEmbedProvider {
+            cache_location: cache_location
+        }
+    }
+
+    pub fn download(&self, model: &AIModel) {
```
deven96, line 60: Nice nice.... we can have a `ModelProviderError` enum that derives `thiserror::Error` and encompasses all the possible errors that could occur during download... and then the function signature becomes

```rust
pub fn download(&self, model:..) -> Result<(), ModelProviderError>;
```

ahnlich/types/src/ai/mod.rs
```diff
@@ -12,8 +12,14 @@ use crate::keyval::StoreInput;
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum AIModel {
     // Image model
-    DALLE3,
+    Dalle3,
```
deven96, line 15: Would leave a TODO here to get rid of those fake `Llama3` and `Dalle3` entries that we noobs added afterwards... now that we have actual models

ahnlich/ai/src/cli/server.rs
```diff
@@ -92,6 +107,11 @@ pub struct AIProxyConfig {
     /// List of ai models to support in your aiproxy stores
     #[arg(long, required(true), value_delimiter = ',')]
     pub(crate) supported_models: Vec<SupportedModels>,
+
+    /// Directory path for storing the model artifacts
+    #[arg(long, default_value_os_t =
+    DEFAULT_CONFIG.get_or_init(AIProxyConfig::default).model_cache_location.clone())]
```
deven96, line 113: I like this trick... could you please do it for other defaults? so we reduce our duplication

deven96 (changes requested):
Added some comments

ahnlich/ai/src/engine/ai/providers/fastembed.rs
```diff
@@ -0,0 +1,71 @@
...
+        let model_type = FastEmbedModel::try_from(model).expect("Model not found");
+
+        if let FastEmbedModel::Text(model_type) = model_type {
+            let cache = Cache::new(self.cache_location.clone());
```
deven96, line 64: Do we need to do any checks if the model already exists in cache and confirm using the hash? Or does `Cache already handle that internally?

ahnlich/types/src/ai/mod.rs
HAKSOAT: I intend removing it when the PR is done.

ahnlich/ai/src/engine/ai/models.rs
HAKSOAT: If we do that, the hash will have to be hardcoded, right?

The challenge so far is, I don't fully understand how Huggingface currently do their validation. I'm just using the functionality as is.

ahnlich/ai/src/engine/ai/providers/fastembed.rs
HAKSOAT: Cool. I don't know how to derive yet, but I will learn how it works.

ahnlich/ai/src/cli/server.rs
HAKSOAT: Makes sense. I'll do that.

ahnlich/ai/src/engine/ai/providers/fastembed.rs
HAKSOAT: They do checks to see if the model exists here:

https://github.com/huggingface/hf-hub/blob/a66bdcf2f2664a55ab3a1415184c7aebd8ac4ed5/src/api/sync.rs#L472-L478

My concern is that that check does not look like any kind of hashing involved, from looking deeper, there are hashed values but it just looks like checking if the file exists (the file is named by the hash value).

ahnlich/ai/src/engine/ai/providers/providers.rs
HAKSOAT: Something that may be more useful that we can keep in mind is the user providing their own models.

ahnlich/ai/src/engine/ai/providers/fastembed.rs
deven96: Then I think we should use the checksums for ensuring correctness, else I predict weird things in our future

ahnlich/ai/src/engine/ai/models.rs
deven96: Yeah we would hard code the hash but we don't have to show the users that info, we just use it to compare for correctness

ahnlich/ai/src/manager/mod.rs
```diff
@@ -167,27 +160,28 @@ impl Task for ModelThread {
 
 #[derive(Debug)]
 pub struct ModelManager {
-    models: HashMap<SupportedModels, mpsc::Sender<ModelThreadRequest>>,
+    models: HashMap<Model, mpsc::Sender<ModelThreadRequest>>,
```
deven96, line 163: I think we may not need `model.clone()` if we leave this as `SupportedModels` or even `AIModel`. Since we already have the model itself within the `ModelThread` and what we receive via the external facing api `handle_request` is an `AIModel`, we don't need cloning, we can rather use the sparse copyable enums

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -98,6 +138,66 @@ impl Model {
             Model::Image { description, .. } => description.clone(),
         }
     }
+
+    pub fn setup_provider(&mut self, cache_location: PathBuf) {
+        let supported_model: SupportedModels = SupportedModels::from(&*self);
+        match self {
+            Model::Text { provider, .. } => {
```
deven96, line 145: I think you can do one block of 

```rust
match self {
   Model::Text { provider, .. } | Model::Image { provider, ... } => {
       match provider {
       }
   }
}
```

ahnlich/ai/src/cli/server.rs
```diff
@@ -177,6 +213,30 @@ impl From<&SupportedModels> for Model {
...
 }
 
+impl From<&Model> for SupportedModels {
+    fn from(value: &Model) -> Self {
+        let supported_model = match value {
+            Model::Text { name, .. } => {
+                match name.as_str() {
+                    "all-MiniLM-L6-v2" => SupportedModels::AllMiniLML6V2,
+                    "all-MiniLM-L12-v2" => SupportedModels::AllMiniLML12V2,
+                    "bge-base-en-v1.5" => SupportedModels::BGEBaseEnV15,
```
deven96, line 223: I think this doesn't match with the name assigned in `model.rs`.... but I think generally if only there was a way to define names in just one single place as opposed to the 3 places we have right now i.e (clap-name, this conversion and AI into Model conversion)

Particularly because they're just strings and not types and so we can skip on things like equality or forgetting usecases by mistake

deven96: To solve this I have an inkling of an approach. We can do it like this

We keep the existing `SupportedModels` with the `clap(name)` derivations. but then we implement the enum and define a kind of Display name

```rust
enum SupportedModels {
     #[clap(name = "all-mini-lm-l6-v2")]
    AllMiniLML6V2,
}

impl std::fmt::Display for SupportedModels  {
  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let name = match self {
         AllMiniLML6V2 =>. "all-MiniLM-L6-V2",         
        _ => ...
        };
        write!(f, "{}", name)
    }}
```

This way, instead of storing a name inside `Model`, we can store a `SupportedModels` and turn it into a string wherever else we need it.... it would remove string matching from this `From` statement and also in `list_supported_models`, we can get name from calling `supported_model.to_string()`

ahnlich/ai/src/manager/mod.rs
```diff
@@ -197,8 +191,8 @@ impl ModelManager {
         inputs: Vec<StoreInput>,
         preprocess_action: PreprocessAction,
     ) -> Result<Vec<StoreKey>, AIProxyError> {
-        let supported = model.into();
-        if let Some(sender) = self.models.get(&supported) {
+        let model = model.into();
```
deven96, line 194: We wouldn't need `AIModel` to convert into `Model` anymore, it can just convert to `SupportedModel`, and that would be used to retrieve the model thread mpsc sender that we need

ahnlich/ai/src/engine/ai/providers/fastembed.rs
deven96: To be left for a different PR

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+        Self {
+            cache_location: None,
+            cache_location_extension: PathBuf::from("huggingface"),
+            supported_models: None,
+            model: None,
+        }
+    }
+
+    pub fn normalize(v: &[f32]) -> Vec<f32> {
+        let norm = (v.iter().map(|val| val * val).sum::<f32>()).sqrt();
```
deven96, line 83: We can use rayon `par_iter` here

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+        let epsilon = 1e-12;
+
+        // We add the super-small epsilon to avoid dividing by zero
+        v.iter().map(|&val| val / (norm + epsilon)).collect()
```
deven96, line 87: Rayon here as well

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+        self.cache_location = Some(location.join(self.cache_location_extension.clone()));
+        self
+    }
+
+    fn set_model(&mut self, model: &SupportedModels) -> &mut Self {
+        self.supported_models = Some(*model);
+        self
+    }
+
+    fn load_model(&mut self) -> &mut Self {
```
deven96, line 102: `load_model` looks very fragile as there's a lot of unwraps.... can we instead pipe these upwards?

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+                    input_param,
+                    output_param,
+                    session: Some(session),
+                }));
+            }
+        }
+        self
+    }
+
+    fn get_model(&self) {
```
deven96, line 156: Same as above.... we would all feel safer if we only unwrap whenever we are absolutely certain it can't break (hence `.expect` or we want it to break the entire flow

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::default::Default;
+use std::fmt;
+use std::path::{Path, PathBuf};
+use std::thread::available_parallelism;
+
+#[derive(Default)]
+pub struct ORTProvider {
+    cache_location: Option<PathBuf>,
+    cache_location_extension: PathBuf,
+    supported_models: Option<SupportedModels>,
```
deven96, line 20: A bit confused why this would be optional

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            let image = match input {
+                ModelInput::Image(image) => image.clone(),
+                _ => {
+                    let store_input_type: AIStoreInputType = input.into();
+                    let index_model_repr: Model = (&self.supported_models.expect(
+                        &AIProxyError::AIModelNotInitialized.to_string())).into();
+                    match action_type {
+                        InputAction::Query => {
+                            panic!(
```
deven96, line 207: There's no need to panic if we are already passing the error upwards... we can just do `return Err(AIProxyError::StoreSetTypeMismatchError(_))`

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+                }
+            };
+
+            let image_arr = image.get_array();
+            let mut image_arr = image_arr.mapv(f32::from);
+            image_arr.swap_axes(1, 2);
+            image_arr.swap_axes(0, 1);
+
+            let array = vec![image_arr.view()];
+            let pixel_values_array = ndarray::stack(ndarray::Axis(0), &array).unwrap();
```
deven96, line 234: Can we run this up the error chain as well?

ahnlich/ai/src/error.rs
```diff
@@ -24,8 +24,8 @@ pub enum AIProxyError {
     UnexpectedDBResponse(String),
     #[error("Cannot Query Using Input. Store expects [{store_query_model_type}], but input type [{storeinput_type}] was provided")]
     StoreQueryTypeMismatchError {
-        store_query_model_type: AIStoreInputType,
-        storeinput_type: AIStoreInputType,
+        store_query_model_type: String,
```
deven96, line 27: I think we can keep these as `AIStoreInputType` not `String`... so we know it can only happen with those types

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -117,8 +241,146 @@ impl ModelInfo {
...
+    Image(ImageArray),
+}
+
+#[derive(Debug, Clone, PartialEq, Hash, Eq)]
+pub struct ImageArray {
+    array: Array<u8, Ix3>,
+    bytes: Vec<u8>,
+}
+
+impl ImageArray {
```
deven96, line 277: Nice work with this internal abstraction.... makes a lot of sense to reason about it this way

deven96 (changes requested):
Added some comments

ahnlich/ai/src/cli/server.rs
```diff
@@ -29,47 +43,105 @@ pub enum Commands {
...
 }
 
+static DEFAULT_CONFIG: OnceLock<AIProxyConfig> = OnceLock::new();
+
 #[derive(Args, Debug, Clone)]
 pub struct AIProxyConfig {
     /// Ahnlich AI proxy port
-    #[arg(long, default_value_t = 1370)]
+    #[arg(long, default_value_t =
+    DEFAULT_CONFIG.get_or_init(AIProxyConfig::default).port.clone())]
```
deven96, line 52: Some of these things have been moved to a shared config which we then called `common` on main and we share it between DB and AI

ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -0,0 +1,266 @@
...
+                input_param,
+                output_param,
+                ..
+            }) => {
+                let model_repo = api.model(repo_name.clone());
+                let model_file_reference = model_repo
+                    .get(&weights_file)
+                    .unwrap_or_else(|_| panic!("Failed to retrieve {} ", weights_file));
+                let executioners: Vec<ExecutionProviderDispatch> = Default::default();
+                let session = Session::builder()
```
deven96, line 134: We can make Session builder errors its own variant which we move upwards with `?`

ahnlich/ai/src/engine/ai/providers/ort.rs
deven96: Also `action_type` doesn't really matter here as both variants return the same error

ahnlich/ai/src/cli/server.rs
HAKSOAT: Yes, that was also refactored.

ahnlich/ai/src/engine/ai/providers/ort.rs
HAKSOAT: SupportedModels shouldn't be optional, but cache_location is optional because it is usually set when `setup_provider` is called through `Model`. The location comes from the user's config.

ahnlich/ai/src/engine/ai/providers/ort.rs
HAKSOAT: Valid point. I agree.

ahnlich/ai/src/engine/ai/providers/ort.rs
HAKSOAT: The variants are: `StoreQueryTypeMismatchError` and `StoreSetTypeMismatchError`, are they the same?

ahnlich/ai/src/error.rs
HAKSOAT: Yeah, I found that confusing. I changed to String because `StoreSetTypeMismatchError` uses String instead of AIStoreInputType, which is inconsistent for similar errors.

deven96: Yeah points to the fact that I need to unify both really and switch them from Strings

deven96 (approved):
LGTM

PR #94, opened 2024-09-02
Ahnlich AI and DB Release Setup.

### LINKED ISSUE #89: AI Proxy and DB Version

Iamdavidonuh:
Closes #89 

When a release is created, we do the following:

- Build Ahnlich binary depending on the tag made(AI or DB) across multiple platforms
-  Build and Deploy docker images to GHCR

During compilation and testing for Windows, found and fixed an error that stopped a socket from being run in non-blocking. This issue caused the connection pool library used by our Python client to error.
 Error: `[Win 100035]: A non-blocking socket operation could not be completed immediately`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,8 +119,10 @@ jobs:
             run: |
               if [ "${{ needs.build_binaries_and_publish.outputs.bin_name }}" == "ahnlich-db" ]; then
                 echo "DEFAULT_PORT=1369" >> $GITHUB_ENV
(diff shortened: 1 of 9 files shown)


deven96 (approved):
LGTM

Get rid of existing tags, binaries and images pushed

deven96 (approved):

PR #98, opened 2024-09-09
Use Model Manager Within Store Handler

Iamdavidonuh:
This PR does the following:

1. Moves Preprocessing of store inputs into `ModelThread`
2. Passes a reference of `ModelManager` to the `Storehandler` method `set` to convert  `StoreInput`s to `StoreKey`s


deven96 (approved):
LGTM

Issue #100, opened 2024-09-09
Fallible collections


deven96:
Vectors, Hashmaps, e.t.c all implicitly fail whenever allocations fail. We are already capping the global allocations using `cap`, fallible collections would then ensure we don't panic but rather surface the allocation errors

deven96:
It wouldn't be practical or possible to make every datastructure we use fallible, so how about we wrap critical client parts in an `std::panic::catch_unwind` to ensure that if we ever do massive allocations, we are able to report errors from there back to the client

PR #103, opened 2024-09-11
WIP: Moving vectors to fallible vectors

deven96:


Iamdavidonuh (approved):

deven96:
Does a bit of #100 

What's left is swapping out the places where we did `.collect()`, and then doing the process for `HashMap` as well. Doesn't seem like `HashSet` has any entry in fallible collections

PR #104, opened 2024-09-15
Swap our ModelManager Hashmap for Moka Cache

Iamdavidonuh:
This PR allows dropping model threads after some defined idle time in aiproxyconfig.
- Swapped out Hashmap for moka cache, to set TTI
- Added tests to show a model thread was recreated after idle time had passed

### DIFF
--- a/ahnlich/ai/src/cli/server.rs
+++ b/ahnlich/ai/src/cli/server.rs
@@ -94,6 +94,12 @@ pub struct AIProxyConfig {
     /// List of ai models to support in your aiproxy stores
     #[arg(long, required(true), value_delimiter = ',')]
     pub(crate) supported_models: Vec<SupportedModels>,
+
+    /// AI Model Idle Time in seconds
+    /// Time in seconds for an unused/idle supported model to be frozen
+    /// Defaults to 5 mins
+    #[arg(long, default_value_t = 60 * 5)]
+    pub(crate) ai_model_idle_time: u64,
 }
 
 impl Default for AIProxyConfig {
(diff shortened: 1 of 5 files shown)


ahnlich/ai/src/cli/server.rs
```diff
@@ -94,6 +94,12 @@ pub struct AIProxyConfig {
     /// List of ai models to support in your aiproxy stores
     #[arg(long, required(true), value_delimiter = ',')]
     pub(crate) supported_models: Vec<SupportedModels>,
+
+    /// AI Model Idle Time in seconds
+    /// Time in seconds for an unused/idle supported model to be frozen
```
deven96, line 99: Not frozen. It gets dropped completely. Also add that calling it after drop reinitializes the model from scratch

ahnlich/ai/src/manager/mod.rs
```diff
@@ -167,27 +171,69 @@ impl Task for ModelThread {
...
+        Ok(Self {
+            models,
+            supported_models: supported_models.to_vec(),
+            task_manager,
+            has_recreated_models: Arc::new(AtomicBool::new(false)),
+        })
+    }
+
+    // QUESTION? Should we hard code the buffer size or add it to config?
+    #[tracing::instrument]
```
deven96, line 206: You probably don't want to trace TaskManager

ahnlich/ai/src/manager/mod.rs
```diff
@@ -167,27 +171,69 @@ impl Task for ModelThread {
...
+    async fn create_request_sender(
+        supported_model: &SupportedModels,
+        task_manager: &TaskManager,
```
deven96, line 209: You can unify this to just take self if you create an empty cache in ModelManager::new() and then proceed to call try_get_with for all of your supported models

deven96 (approved):
LGTM!

PR #108, opened 2024-09-19
Optimizations: StoreInputs Validation and Data Prep

Iamdavidonuh:
As part of #106,  We want to speed up the processing of store inputs. Here we spawn tokio threads, utilizing the `FuturesOrdered` struct, which imposes `FIFO` thereby ensuring Order. This helps as we want to fail quickly if we get a validation error on a spawned thread.

PR #109, opened 2024-09-20
ThreadPool for StoreInput Validations

Iamdavidonuh:
As part of #108 .

In this PR, I'm creating a ThreadPool that uses the task manager to N number of tasks for store_inputs validation.

Potential Changes:

- Maybe we move the thread pool into the task manager and have certain use cases define their function signature?


Iamdavidonuh:
Closing this, as it was just for POC

Issue #106, opened 2024-09-16
Potential Performance Improvements


Iamdavidonuh:
- [x] `validate_and_prepare_store_data` is running in a loop
- [x]  `handle_request` needs to be more granular, and current runs in a loop.
- [x] `ahnlich_db::engine::store:add` takes time
- [x] Setup SIMD for linear similarity algorithms (#187)

PR #110, opened 2024-09-20
Optimizations: StoreInputs Validation and Data Prep 

Iamdavidonuh:
As part of #106, We want to speed up the processing of store inputs. Here we spawn a fixed amount of Tokio threads. This helps as we want to fail quickly if we get a validation error on a spawned thread.


deven96 (approved):
LGTM

PR #134, opened 2024-10-24
Added batch inferencing

HAKSOAT:
This pull request contains the following major changes:

- Batch inferencing for the FastEmbed and ORT providers
- Batch preprocessing for the images and the text
- Change in the interface for the Provider trait `run_inference` for more efficiency

### DIFF
--- a/ahnlich/ai/src/engine/ai/models.rs
+++ b/ahnlich/ai/src/engine/ai/models.rs
@@ -9,7 +9,7 @@ use ahnlich_types::{
     keyval::{StoreInput, StoreKey},
 };
 use image::{GenericImageView, ImageReader};
-use ndarray::{Array, Array1, Ix3};
+use ndarray::{Array, Ix3};
 use nonzero_ext::nonzero;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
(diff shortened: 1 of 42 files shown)


ahnlich/ai/src/engine/ai/providers/fastembed.rs
```diff
@@ -182,36 +184,35 @@ impl ProviderTrait for FastEmbedProvider {
...
-                            return Err(AIProxyError::AIModelNotInitialized);
-                        };
-                        let index_model_repr: Model = (&index_model_repr).into();
-                        return Err(AIProxyError::StoreTypeMismatchError {
-                            action: *action_type,
-                            index_model_type: index_model_repr.input_type(),
-                            storeinput_type: store_input_type,
-                        });
-                    }
+    fn run_inference(&self, inputs: &[ModelInput], action_type: &InputAction) -> Result<Vec<Vec<f32>>, AIProxyError> {
```
deven96, line 187: Can `run_inference` return the `Vec<StoreKey>`? so that higher up callers no longer have to iterate over the vec just to cast to another type

ahnlich/ai/src/engine/ai/models.rs
```diff
@@ -120,19 +120,18 @@ impl Model {
...
-                let embedding = provider.run_inference(storeinput, action_type)?;
-                Ok(StoreKey(<Array1<f32>>::from(embedding)))
-            }
-        }
+    ) -> Result<Vec<StoreKey>, AIProxyError> {
+        let embeddings = match &self.provider {
+            ModelProviders::FastEmbed(provider) => provider.run_inference(storeinput, action_type)?,
+            ModelProviders::ORT(provider) => provider.run_inference(storeinput, action_type)?
+        };
+        let store_keys = embeddings
```
deven96, line 130: If we do [this](https://github.com/deven96/ahnlich/pull/134/files#r1815264646) we can get rid of this

deven96 (changes requested):
There shouldn't be any changes in the python lib I believe as protocol wasn't tampered with... so perhaps the changes that show up in Python are old? You can force checkout `sdk/ahnlich-client-py` from main to fix those

HAKSOAT:
> There shouldn't be any changes in the python lib I believe as protocol wasn't tampered with... so perhaps the changes that show up in Python are old? You can force checkout `sdk/ahnlich-client-py` from main to fix those

The changes are cosmetic, it is because I ran typegen, but I will undo them.

deven96 (approved):

Issue #105, opened 2024-09-16
Preprocess action for GetSimN on ai-proxy


deven96:
In the same vein of preprocess action for ai-proxy Set command

PR #135, opened 2024-10-24
AI Proxy request and response Fixes

### LINKED ISSUE #131: AI store without storing originals
Right now by default we store originals passed to AI within metadata that we store to DB. However that could lead to duplication of information as its very possible that the original is stored elsewhere

By including a new argument to AI CREATESTORE called `store_original: bool`, we can then change AI response from being `Vec<(StoreKey, StoreValue, Similarity)>` to being `Vec<(Option<StoreKey>, StoreValue, Similarity>`

This would then be really light in potential mobile use cases... only storing the embeddings and metadata

To do this we would have to
- [ ] Modify CREATESTORE query, GETSIMN response, and GETPRED response to account for `store_original` and Optional in protocol
- [ ] Fix CREATESTORE within DSL
- [ ] Fix CREATESTORE, GETSIMN, GETPRED within Rust and Python client libraries

deven96:
- Including `error_if_exists` to `CREATESTORE` args to solve part of #105 
- Including `store_original` to `CREATESTORE` args to begin the work of #131 
- Modifying the responses of `AIServerResponse::Get` and `AIServerResponse::GetSimN` to reflect the fact that originals might be `None`. This finally closes #131


ahnlich/ai/src/server/task.rs
```diff
@@ -153,27 +155,35 @@ impl AhnlichProtocol for AIProxyTask {
...
-                    {
-                        Ok(res) => {
-                            if let ServerResponse::Del(num) = res {
-                                Ok(AIServerResponse::Del(num))
-                            } else {
-                                Err(AIProxyError::UnexpectedDBResponse(format!("{:?}", res))
-                                    .to_string())
+                    match self.store_handler.store_original(store.clone()) {
+                        Err(err) => Err(err.to_string()),
+                        Ok(false) => Err(AIProxyError::DelKeyError.to_string()),
```
HAKSOAT, line 160: If we can't delete key for cases when `store_original` is `false`, how do we get to delete such inputs? Will they be stuck in the db forever? Or can we currently then delete via their metadata?

ahnlich/ai/src/engine/store.rs
```diff
@@ -203,11 +212,18 @@ impl AIStoreHandler {
...
+                    return Err(AIProxyError::ReservedError(metadata_key.to_string()));
+                }
+                let metadata_value: MetadataValue = store_input.clone().into();
+                store_value.insert(metadata_key.clone(), metadata_value.clone());
+                delete_hashset.insert(metadata_value);
+            }
             output.try_push((store_input, store_value))?;
-            delete_hashset.insert(metadata_value);
         }
+        let delete_hashset = (store_original).then_some(delete_hashset);
```
HAKSOAT, line 226: Not storing originals also means we can't update values.

If this is right, instead of removing the original completely, is it possible for us to hash the original, so we can have something to work with instead of removing completely.

deven96: I think we can justify not being able to update the original.... the hash itself can be pretty large as well and still costs some cycles

One thing we can offer users is to document that when not storing the original, the upside is no duplication but the downside we are unable to update that particular entry and would need them to first issue an appropriate `DelPred` on their end using any appropriate metadata set

ahnlich/ai/src/server/task.rs
deven96: We as the server can't because we don't track any other metadata entry... but the user can

deven96: For instance if the user decides to introduce a metadatavalue entry of (key, uuid) that's always guaranteed to be the same per entry then yes they can delete that ... perhaps that points to having a more atomic `SetPred` command where you can set that key where the predicate suffices

HAKSOAT: What command will they use to delete in that case?

deven96: They can use `DelPred`

PR #137, opened 2024-10-27
Storing f32 to later avoid allocation a Vec of inputs

deven96:


ahnlich/ai/src/engine/ai/providers/ort.rs
```diff
@@ -252,15 +256,6 @@ impl ProviderTrait for ORTProvider {
...
             });
         }
-
-        let batch_size = 16;
-        let store_keys = image_inputs
-            .chunks(batch_size)
-            .try_fold(Vec::new(), |mut accumulator, batch_inputs|{
-                accumulator.extend(self.batch_inference(batch_inputs)?);
-                Ok(accumulator)
-            });
```
HAKSOAT, line 262: I don't think we should remove the use of batches.

HAKSOAT: If for example we have 1000 inputs, it is better to run them in batches of say 32 for example, than doing all 1000 at once.

HAKSOAT (approved):

PR #141, opened 2024-11-08
Book search example

SLKhadeeja:


deven96:
Reviews are already clear... will open as full PR once implemented

deven96 (changes requested):
Formatting still fails but everything else LGTM

deven96 (approved):

PR #151, opened 2024-11-30
Trim builder methods for GetSimN, add default to condition

Iamdavidonuh:


examples/rust/image-search/src/main.rs
```diff
@@ -9,7 +9,6 @@ use ahnlich_types::{
     ai::{AIModel, AIServerResponse, PreprocessAction},
     keyval::{StoreInput, StoreName},
     metadata::{MetadataKey, MetadataValue},
-    similarity::Algorithm,
```
deven96, line 12: Could we also make the default processing to be ModelProcessing?

Iamdavidonuh: Sure why not

deven96 (approved):

PR #190, opened 2025-01-27
Slight Resizing Optimizations  for `preprocess_images`🏃‍♂️ 

deven96:
I was testing out various CUDA runtimes using Google Collab such as
* A100 GPU
* L4 GPU
* T4 GPU
This led me to understand that our current `preprocess_images` was quite slow accross the board and inference speed was running in tens of milliseconds. This was run using the very simple `Image2ImageDemo` test in `demo_embed.py`. 

So I cloned the repo, installed `jaeger` binary and `ngrok` on the machine as it's not possible seemingly to be able to run the docker service on the collab terminal
Screenshot 2025-01-27 at 01 20 41

Poking around, I figured out the way we preprocessed held onto quite a bit of redundant data i.e array & image and we had a bit of cloning going on. It was also constrained by the very slow resizing method of our current `image` crate. It's an amazing crate for broad decoding and it's `crop` function also isn't so bad ... but when it comes to resizing it was lagging behind. 

Tbh I could have done a bit of poking around myself to figure out why that was the case, but it turns out there was an amazing crate called [`fast_image_resize` ](https://docs.rs/fast_image_resize) that has
- Reasonable interoperability with the `image` crate
- Extremely fast image resizing that leverages SIMD instructions and rayon
Screenshot 2025-01-27 at 04 58 28

Next things to test out and optimize are 
- `preprocess_store_input`: This now shows up as the new bottleneck for image processing despite having rayon capabilities
- `preprocess_text`: I'm positive some red flags would likely pop up here as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,11 @@ services:
 
 ```
 
+### Execution Providers (Ahnlich AI)
+
+`CUDA`: Only supports >= CUDAv12 and might need to `sudo apt install libcudnn9-dev-cuda-12`
+`CoreML (Apple)`: Not advised for NLP models due to often large dimensionality.
+
 ### Contributing
 
 View [contribution guide](CONTRIBUTING.md)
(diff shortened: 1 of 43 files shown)


HAKSOAT (approved):
Thanks for introducing fast_image_resize. Good stuff.

PR #195, opened 2025-02-06
Ahnlich Protocol Swap Spec

Iamdavidonuh:


protos/version.proto
deven96, line 1: Perhaps add a README.md to the new `protos` folder explaining the structure and potential services

ahnlich/grpc_types/build.rs
```diff
@@ -0,0 +1,123 @@
+use std::{
+    collections::HashSet,
+    io::{Result, Write},
+    path::PathBuf,
+};
+use walkdir::WalkDir;
+
```
deven96, line 7: Add a `TODO` or `FIXME` here that shows how we aim to replace `types` with `grpc_types` i.e turn the latter into the former

protos/version.proto
deven96: The README can then contain how the protos are used to generate types for `rust`, `python`, e.t.c .... with rust automatically happening with any cargo command and python happening on ??

protos/keyval.proto
```diff
@@ -0,0 +1,41 @@
...
+message StoreEntry {
+    StoreKey key = 1;
+    StoreValue value = 2;
+}
+
+
+message StoreValue {
+    map<string, metadata.MetadataValue> value = 1;
+
+}
```
deven96, line 41: We still need to map out services for AI and DB

deven96 (changes requested):
Add protoc to pipelines that will need it

deven96:
An awesome library to use for Python is [betterproto](https://github.com/danielgtaylor/python-betterproto)

protos/version.proto
deven96: For python we can leverage pyproject.toml custom build script with betterproto like we did with cargo in rust

protos/services/db_service.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package services.db_service;
+
+import "db/query.proto";
+import "db/server.proto";
+import "db/pipeline.proto";
+
+service DBService {
+    rpc CreateStore (db.query.CreateStore) returns (db.server.Unit);
```
deven96, line 10: Optional tracing id should be passed to every rpc call

deven96: We've spoken that it would be a grpc header instead of getting inserted into the messages

deven96 (approved):

PR #204, opened 2025-07-04
Go grpc client

derhnyel:
The go client grpc


sdk/ahnlich-client-go/tests/utils_test.go
```diff
@@ -0,0 +1,13 @@
+package ahnlichgotest
```
deven96, line 1: Do we need this file?

derhnyel: I don't think so; was using it for sanity check

deven96 (approved):

PR #229, opened 2025-07-08
Fix persistence with custom serialize deserialize

deven96:
Persistence was broken with the new grpc types due to `MetadataValue` is a key within `PredicateIndex` so I 
- Modified the persistence test to ensure it includes a key that creates an entry of `PredicateIndex`
- Implemented custom serialize deserialize for `MetadataValue` making use of [ascii85](https://en.wikipedia.org/wiki/Ascii85) standard which should give us a string representation while being more compact than base64
- Removed old custom_metadata_map implementation as `papaya` supports serde by default so we don't need that any longer


Iamdavidonuh (approved):
LGTM

PR #246, opened 2025-08-28
added basic bones for client library structure

SLKhadeeja:


deven96:
Seems we are already past this

PR #247, opened 2025-09-08
feat: decouple DB

jimezesinachi:
Step 1: add new data structures to protobufs, updated AIProxyServer to make `db_client` optional & updated `create_store` in AIService

Step 2: updated protos and updated the rest of the handlers

Step 3: added `AiService::convert_store_input_to_embeddings()` business logic

Step 4: updated `ModelManager::handle_request()` to use Arc for the StoreInputs

Step 5: added tests for new code

### DIFF
--- a/ahnlich/ai/src/manager/mod.rs
+++ b/ahnlich/ai/src/manager/mod.rs
@@ -312,7 +312,7 @@ impl ModelManager {
 
         let (response_tx, response_rx) = oneshot::channel();
         let request = ModelThreadRequest {
-            inputs: inputs,
+            inputs,
             response: response_tx,
             preprocess_action,
             action_type,
--- a/ahnlich/ai/src/server/handler.rs
+++ b/ahnlich/ai/src/server/handler.rs
@@ -1040,7 +1040,7 @@ impl AIProxyServer {
             client_handler,
             store_handler: Arc::new(store_handler),
             config,
-            db_client: db_client,
+            db_client,
             task_manager,
             model_manager: Arc::new(model_manager),
         })


ahnlich/ai/src/manager/mod.rs
```diff
@@ -312,7 +312,7 @@ impl ModelManager {
 
         let (response_tx, response_rx) = oneshot::channel();
         let request = ModelThreadRequest {
-            inputs,
+            inputs: inputs.as_ref().clone(),
```
deven96, line 315: Do we need `.as_ref()`.. should be able to clone the arc itself with `Arc::clone`

deven96: Might need modifying ModelThreadInput though

ahnlich/ai/src/server/handler.rs
```diff
@@ -574,6 +621,65 @@ impl AiService for AIProxyServer {
...
+        let store_keys = ModelManager::handle_request(
+            &self.model_manager,
+            &ai_model,
+            Arc::clone(&inputs),
+            preprocess_action,
+            InputAction::Index,
+            None,
+        )
+        .await?;
+
```
deven96, line 668: We probably need to ensure first that they are of the same length before we zip... unless we already do that somewhere within `handle_request` but I doubt that

ahnlich/ai/src/tests/aiproxy_test.rs
```diff
@@ -623,6 +624,179 @@ async fn test_ai_proxy_get_sim_n_succeeds() {
     );
 }
 
+#[tokio::test]
+async fn test_convert_store_input_to_embeddings() {
```
deven96, line 628: Nice idea on the tests... I like it

We can probably parameterize the models with with [test_case](https://docs.rs/test-case/latest/test_case/) crate as a dev dependency for ai

deven96: Would then mean we can remove the for loop

ahnlich/ai/src/server/handler.rs
```diff
@@ -574,6 +621,65 @@ impl AiService for AIProxyServer {
         }))
     }
 
+    #[tracing::instrument(skip_all)]
+    async fn convert_store_input_to_embeddings(
```
deven96, line 625: Would need to add support for this to the rust, golang and python AI clients respectively

You've generated the types already for Rust and Python , but yet to generate for Golang

But for Rust, there's a bit of boilerplate (I should really turn that into a macro) within`ahnlich/client/src/ai.rs` to neatly wrap it up for Rust clients pipeline and non-pipeline usage

For Python I think a test would be helpful to confirm this works

For Golang, there is a Makefile in the directory to regenerate and probably also a test

ahnlich/ai/src/server/handler.rs
```diff
@@ -90,7 +95,7 @@ pub struct AIProxyServer {
     client_handler: Arc<ClientHandler>,
     store_handler: Arc<AIStoreHandler>,
     task_manager: Arc<TaskManager>,
-    db_client: Arc<DbClient>,
+    db_client: Option<Arc<DbClient>>,
```
deven96, line 98: So now we need a CLI argument such as `--without-db` with which to spin AI proxy up with for which db_client would be `None`.... the default should remain as is i.e it seeks to initialize a connection to DB with `--db-host` and `--db-port` ... both argument groups should be mutually exclusive i.e `--without-db` and `--db-*` arguments cannot be passed together perhaps with something like [this](https://rust.code-maven.com/clap/clap-mutually-exclusive-with-group)

deven96 (changes requested):

ahnlich/ai/src/tests/aiproxy_test.rs
```diff
@@ -623,6 +624,179 @@ async fn test_ai_proxy_get_sim_n_succeeds() {
...
+        ];
+
+        let pipelined_request = ai_pipeline::AiRequestPipeline { queries };
+
+        let _ = client
+            .pipeline(tonic::Request::new(pipelined_request))
+            .await
+            .expect("Failed to send pipeline request");
+
+        let expected_entries = ai_response_types::StoreInputToEmbeddingsList {
```
deven96, line 759: If we switch to the assert like below we don't really need this then as we just need to clone each store_input to check that they exist in the response

ahnlich/ai/src/tests/aiproxy_test.rs
```diff
@@ -623,6 +624,179 @@ async fn test_ai_proxy_get_sim_n_succeeds() {
...
+        };
+
+        let response = client
+            .convert_store_input_to_embeddings(tonic::Request::new(query))
+            .await
+            .expect("Failed to convert store input to embeddings");
+
+        let response_entries = response.into_inner().values;
+
+        assert_eq!(response_entries.len(), expected_entries.values.len());
```
deven96, line 787: This would then just be `assert_eq!(store_inputs.len(), response_entries.len())`

ahnlich/ai/src/manager/mod.rs
deven96: I'd let you decide.... if at it's very root, we cannot ship the `Arc<Vec<StoreInput>>` to the models for inference without cloning, then perhaps we just get rid of the `Arc` all together and clone .... but feel free to change it all the way down if it works

It won't affect the SDKs at all as this is internal to just AI proxy

ahnlich/ai/src/cli/server.rs
```diff
@@ -301,3 +309,53 @@ impl SupportedModelArgs {
...
+
+#[cfg(test)]
+mod tests {
+    use regex::Regex;
+    use std::process;
+    use test_case::test_case;
+
+    #[test_case("--db-host", Some("127.0.0.1"), Some("<DB_HOST>"); "with_db_host_option_passed")]
+    #[test_case("--db-port", Some("1369"), Some("<DB_PORT>"); "with_db_port_option_passed")]
+    #[test_case("--db-https", None, None; "with_db_https_option_passed")]
```
deven96, line 321: Nice call on the test but it would then seem even "db-https" should be unusable with --without-db then? i.e no point setting that if you've said without db

Would also ellminate all the options and just be `&str`, `&str`, `&str`

deven96: Ah okay I see what you mean... still conficts but only with env var names and placeholders

deven96 (approved):
Looks like we finally there

Upon merge we can talk about if we want it in the CLI

Issue #271, opened 2025-10-28
Horizontal scalability for ahnlich


deven96:
Spec: TBD

Currently ahnlich only has the ability to run one instance of it. Upon completion of this initiative, that should change... allowing for horizontal scalability. First thing that comes to mind is sharding vs replication

PR #276, opened 2025-11-17
draft: RFC for `Horizontal scalability for ahnlich`

jimezesinachi:
Issue - [Horizontal scalability for ahnlich](https://github.com/deven96/ahnlich/issues/271)


ahnlich/replication-rfc.md
```diff
@@ -0,0 +1,41 @@
+# Ahnlich replication RFC
+
+The first thing we want to do is to identify the components we need to implement.
+
+They are (NOTE: these names are tentative):
+
+- `LogStore` - This is where the logs from the Raft cluster activities will be stored. Here is an in-memory impl from the openraft guys that they used in their example: <https://github.com/databendlabs/openraft/blob/4f0fd5fa034413d2f367306da4a0016f7603fb7e/examples/mem-log/src/log_store.rs>, and I believe we can easily co-opt this into a write-to-disk-log file service, or any other log storage we settle on.
+
+- `StateMachineStore` - This is where the last known state (snapshot) is stored and read from. They have a neat impl here as well: <https://github.com/databendlabs/openraft/blob/4f0fd5fa034413d2f367306da4a0016f7603fb7e/examples/raft-kv-memstore-grpc/src/store/mod.rs>, and I think the bit we need to figure out are where we want to 'store' the state machine.
```
deven96: What is a snapshot in this sense with respect to the state? All the application data in the store in some serializable format?

ahnlich/replication-rfc.md
```diff
@@ -0,0 +1,41 @@
+# Ahnlich replication RFC
+
+The first thing we want to do is to identify the components we need to implement.
+
+They are (NOTE: these names are tentative):
+
+- `LogStore` - This is where the logs from the Raft cluster activities will be stored. Here is an in-memory impl from the openraft guys that they used in their example: <https://github.com/databendlabs/openraft/blob/4f0fd5fa034413d2f367306da4a0016f7603fb7e/examples/mem-log/src/log_store.rs>, and I believe we can easily co-opt this into a write-to-disk-log file service, or any other log storage we settle on.
```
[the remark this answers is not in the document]
jimezesinachi: It's pretty much durability vs speed. On-disk allows for easier recovery from logs and snapshots, but is slower to write and read from. In-memory can still recover, but only in the case that there's at least 1 node to recover from, else if all nodes go down, all data is lost; but it is faster to write to and read from. I think the question for us to answer is which of these guarantees will be more important to users that will be horizontally scaling their ahnlich

ahnlich/replication-rfc.md
```diff
@@ -0,0 +1,41 @@
...
+- Allow for Raft nodes to be created based on some config/command issued via the CLI
+
+- Pass some Ahnlich data to the Raft cluster that it would replicate through out the cluster
+
+- Allow for the cluster to be started and stopped
+
+- Allow for the cluster to be restarted
+
+The next step I think is answering the following questions.
+
```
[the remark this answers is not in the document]
jimezesinachi: Yep, I can see how it should be possible. Their examples have placeholders for where they expect us to do any persistence stuff we want to add, for example, these lines from here: https://github.com/databendlabs/openraft/blob/956d6f6a6c344d63a92b5bebf81ee9051a1aace2/examples/raft-kv-memstore-grpc/src/store/mod.rs#L91, see below:

```rs
// Emulation of storing snapshot locally
{
    let mut current_snapshot = self.current_snapshot.lock().unwrap();
    *current_snapshot = Some(stored);
}
```
so basically we should be able to just plug in some persistence function in there



For the logs, they are saved in memory as a `BTreeMap`:

```rs
pub struct LogStoreInner<C: RaftTypeConfig> {
    /// The last purged log id.
    last_purged_log_id: Option<LogIdOf<C>>,

    /// The Raft log.
    log: BTreeMap<u64, C::Entry>, /// <- Note here

    /// The commit log id.
    committed: Option<LogIdOf<C>>,

    /// The current granted vote.
    vote: Option<VoteOf<C>>,
}
```

And appended with this function:
    
```rs
async fn append<I>(&mut self, entries: I, callback: IOFlushed<C>) -> Result<(), io::Error>
    where I: IntoIterator<Item = C::Entry> {
        // Simple implementation that calls the flush-before-return `append_to_log`.
        for entry in entries {
            self.log.insert(entry.index(), entry); /// <- Note here
        }
        callback.io_completed(Ok(())).await;

        Ok(())
}
```

We should be able to add a persistent function call inside the `append()`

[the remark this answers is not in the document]
deven96: Okay so in the case where a replica goes down and comes back up... does it receive the logs from other nodes in the cluster or does it load from it's possibly outdated logs

I'm guessing it's the former... if so then we may not need to persist RAFT logs on disk and we could stick to instead persisting only the snapshots as we currently do

ahnlich/replication-rfc.md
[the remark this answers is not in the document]
deven96: Ohhhh nice nice .... we can somewhat reuse our existing snapshot in this case then

ahnlich/replication-rfc.md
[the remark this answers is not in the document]
jimezesinachi: > Cuz if it receives the logs from other nodes in the cluster, then on startup it shouldn't absolutely need to recover from disk as it would get the logs and snapshots right?
> 
> Assuming there's atleast other nodes in a good state

Ahh, I understand now

ahnlich/replication-rfc.md
```diff
@@ -0,0 +1,183 @@
...
+Both DB and AI CLIs now include cluster flags:
+
+- `--cluster-enabled`
+- `--raft-node-id`
+- `--raft-addr`
+- `--admin-addr`
+- `--raft-storage` (`memory` or `rocksdb`)
+- `--raft-data-dir` (required when `--raft-storage rocksdb`)
+- `--raft-snapshot-logs`
```
deven96: Could we perhaps get short descriptions for these CLI arguments. I personally would like to know snapshot logs vs storage

ahnlich/db/src/server/handler.rs
```diff
@@ -27,18 +36,40 @@ use utils::allocator::GLOBAL_ALLOCATOR;
...
     config: ServerConfig,
+    raft: Option<Arc<Raft<crate::replication::TypeConfig>>>,
+    raft_registry: Option<NodeRegistry>,
+    request_seq: AtomicU64,
+    pending_join: Option<String>,
+}
+
+impl std::fmt::Debug for Server {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AIProxyServer")
```
deven96: `DBServer` instead of `AIProxyServer`

ahnlich/db/src/server/handler.rs
```diff
@@ -840,13 +1113,22 @@ impl BlockingTask for Server {
         mut self,
         shutdown_signal: std::pin::Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>>,
     ) {
+        let token = self.cancellation_token();
+        let token_clone = token.clone();
+
+        tokio::spawn(async move {
```
deven96: Why was this necessary?

ahnlich/db/src/cli/server.rs
```diff
@@ -18,6 +19,34 @@ pub enum Commands {
...
+    #[arg(
+        long,
+        requires = "cluster_enabled",
+        required_if_eq("raft_storage", "rocksdb")
+    )]
+    pub raft_data_dir: Option<std::path::PathBuf>,
+    /// Snapshot after N logs
+    #[arg(long, default_value_t = 1000, requires = "cluster_enabled")]
+    pub raft_snapshot_logs: u64,
+    /// Join existing cluster via admin addr host:port
```
deven96: Would this be necessary if we already have a specified admin_addr default where we try to first join or then create a cluster?

ahnlich/ai/src/engine/store.rs
```diff
@@ -78,8 +83,20 @@ impl AIStoreHandler {
...
+            .stores
+            .iter(&guard)
+            .map(|(name, _)| name.clone())
+            .collect();
+        for key in keys {
+            let _ = self.stores.remove(&key, &guard);
+        }
+        let snap_guard = stores_snapshot.guard();
+        for (key, value) in stores_snapshot.iter(&snap_guard) {
+            let _ = self.stores.try_insert(key.clone(), value.clone(), &guard);
```
deven96: We should probably not ignore these errors... i.e atleast aggressively error log

deven96 (commented):
So far so good I think this is huge progress already

Particularly liking the test setup to ensure quorum... Don't think those got exercised to the point where a snapshot was triggered though i.e we've had so many writes that we trigger an async snapshot savepoint

ahnlich/ai/src/engine/store.rs
jimezesinachi: Makes sense, I'll get on it

PR #284, opened 2025-12-24
Feat/filewatcher go example

chijiooke:
Adds a filewatcher CLI that watches a directory for PDF file changes and posts normalized file events to an HTTP API. Includes event debouncing for macOS, structured logging, and a clean, extensible architecture for future processing pipelines to examples directory for go lang

### DIFF
--- a/examples/go/filewatcher/README.md
+++ b/examples/go/filewatcher/README.md
@@ -109,7 +109,7 @@ Each detected file event is sent as JSON:
 Example logs:
 
 ```text
-[INFO] Watching directory: /Users/michael/Documents
+[INFO] Watching directory: /Users/chijooke/Documents
 [INFO] FILE_ADDED: invoice.pdf
 [INFO] Posted event to API
 ```
--- a/examples/go/filewatcher/README.md
+++ b/examples/go/filewatcher/README.md
@@ -1,0 +1,172 @@
+
+# Filewatcher
+
+A lightweight Golang CLI tool that monitors a directory for PDF file changes and sends normalized file events to an external API in real time.
+
+Designed to support downstream workflows such as AI analysis, embedding pipelines, and event logging.
+
+---
+
+## Features
+
+* Real-time directory monitoring on macOS
+* PDF-only file filtering
+* Detects file lifecycle events:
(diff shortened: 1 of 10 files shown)


examples/go/filewatcher/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+---
+
+## Features
+
+* Real-time directory monitoring on macOS
+* PDF-only file filtering
+* Detects file lifecycle events:
+
+  * `FILE_ADDED`
```
deven96: Nice work outlining this

I'm curious how file renames are represented. As a `FILE_UPDATED`? or as a `FILE_REMOVED+FILE_ADDED`

examples/go/filewatcher/cmd/root.go
```diff
@@ -0,0 +1,27 @@
...
+	Dir     string
+	APIURL  string
+	Verbose bool
+}
+
+func ParseFlags() (*Config, error) {
+	cfg := &Config{}
+
+	flag.StringVar(&cfg.Dir, "dir", "", "Directory to watch")
+	flag.StringVar(&cfg.APIURL, "api-url", "", "API endpoint")
```
deven96: HTTP API is not a bad choice tbh but given that we expect the watcher to run on the same system as the future components (atleast for this example), how about a unix domain socket path?

examples/go/filewatcher/README.md
```diff
@@ -0,0 +1,172 @@
...
+## Error Handling
+
+* Directory not found → Program exits with a descriptive error
+* API failure → Logged; watcher continues running
+* Permission issues → Logged as warnings
+* Non-PDF files → Ignored
+
+---
+
+## Project Structure
```
deven96: The file watcher is one piece out of the project... perhaps the entire project should be named something like `pdfsearch` and the filewatcher is one binary under pdfsearch that pushes events to the indexer

examples/go/filewatcher/cmd/root.go
deven96: I think this is deprecated now given the structure below

PR #295, opened 2026-02-17
Defer `size_of_store` computation 

### LINKED ISSUE #249: Defer "size of store" computation in `liststore` command
Currently we compute the size of each store when the `liststore` command is issued. Having this value precomputed and stored would speed up the liststore request. Although we'd sacrifice some form of consistency with this approach but i'd say it's a better compromise.

deven96:
Fixes #249 

`LISTSTORE` command triggers the DB to recursively calculate the size of every object in the DB ... as the DB grows this could potentially become very expensive... This patch doesn't solve the algorithmic complexity as that remains `O(N)`. However it just makes it so that we can defer computing the size of a store (and it's nested entries) by creating a new `Task` that can be spawned by our task manager

To do this, we also introduce a per-store write flag that gets updated whenever the store gets modified. This way we have hook points to set our cached atomics and those don't get updated between writes. This also introduced a new hook for starting up tasks for our server i.e `spawn_tasks_before_server` that can return errors if spawn fails and stop server startup

I then reuse that to also unify how `ModelManager` gets access to `TaskManager` and spawns the model threads, unifying task startup as a whole

### DIFF
--- a/ahnlich/ai/src/manager/mod.rs
+++ b/ahnlich/ai/src/manager/mod.rs
@@ -1,5 +1,6 @@
 use std::path::Path;
 use std::sync::Arc;
+use std::sync::OnceLock;
 
 use crate::cli::server::{ModelConfig, SupportedModels};
 use crate::engine::ai::models::{ImageArray, InputAction, ModelResponse};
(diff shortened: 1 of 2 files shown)


ahnlich/utils/src/server.rs
```diff
@@ -51,6 +53,12 @@ pub trait AhnlichServerUtils: BlockingTask + Sized + Send + Sync + 'static + Deb
 
     fn task_manager(&self) -> Arc<TaskManager>;
 
+    /// Spawn background tasks before server starts (e.g., model threads, size calculation)
+    /// Returns error to fail fast if critical initialization fails
+    async fn spawn_tasks_before_server(&self, _task_manager: &Arc<TaskManager>) -> IoResult<()> {
+        Ok(())
```
jimezesinachi: Quick one, is there something else that should be happening inside this function, is this just a placeholder?

PR #325, opened 2026-03-28
RFC: Horizontal scalability spec doc

jimezesinachi:


docs/specs/replication.md
```diff
@@ -0,0 +1,859 @@
...
+### Service Surfaces
+
+Each server process (DB or AI) exposes two gRPC services when clustering is enabled:
+
+| Service                       | Port Flag        | Purpose                                                                                                        |
+|-------------------------------|------------------|----------------------------------------------------------------------------------------------------------------|
+| Public (DbService/AiService)  | `--port`         | Client-facing CRUD and query operations                                                                        |
+| Cluster                       | `--cluster-addr` | Node-to-node Raft RPCs (AppendEntries, Vote, InstallSnapshot) and cluster administration (membership, metrics) |
+
+The Cluster service is **not** exposed in client SDKs. It is internal to the cluster.
```
deven96: Yes its not exposed but I think we need at least some of the APIs in order to be able to really do load balancing from the SDK end

docs/specs/replication.md
```diff
@@ -0,0 +1,859 @@
...
+| `DelKey`                         | `db.query.DelKey`                       | Deletes entries by vector keys                                          |
+| `DelPred`                        | `db.query.DelPred`                      | Deletes entries matching a predicate condition                          |
+| `DropPredIndex`                  | `db.query.DropPredIndex`                | Removes predicate indices from a store                                  |
+| `DropNonLinearAlgorithmIndex`    | `db.query.DropNonLinearAlgorithmIndex`  | Removes non-linear algorithm indices from a store                       |
+| `DropStore`                      | `db.query.DropStore`                    | Deletes an entire store                                                 |
+
+**Not routed through Raft** (read-only operations):
+
+- **Served locally by any node** (follower or leader): `GetKey`, `GetPred`, `GetSimN`, `ListClients`, `InfoServer`, `Ping` — these read from local state and tolerate slightly stale data.
+- **Forwarded to leader** (`ensure_linearizable()`): `ListStores` — consistency-sensitive because users expect to see a store immediately after creating it.
```
deven96, line 222: I think Pipeline calls should also be forwarded to the master node... as there's no way to tell on the protocol level if the pipeline contains a mutation or not

deven96: Either that or if a follower node receives a pipeline request, it should be able to first tell if it contains any mutation, if yes forwards to master, if no it can serve it locally

docs/specs/replication.md
```diff
@@ -0,0 +1,859 @@
...
+
+- **In standalone mode** (no `--cluster-addr`): Standalone persistence operates as before (unchanged behavior).
+- **In cluster mode** (`--cluster-addr` provided): Raft log replication and Raft snapshots are the primary mechanisms for data consistency and recovery across nodes. `Persistence` can optionally remain active as a local backup.
+
+**On startup in cluster mode**: Raft reconciliation (snapshot install from leader + log replay) is authoritative. If `Persistence` is enabled, the local JSON snapshot could be loaded first as a "warm start" to reduce the amount of data the leader needs to transfer, with Raft then reconciling any differences on top. Alternatively, the local snapshot can be ignored entirely and the node can bootstrap purely from the Raft cluster.
+
+> **Open question for the team**: Should we leave `Persistence` enabled in cluster mode as a local backup/warm-start mechanism, or disable it to keep the durability path simple (Raft-only)? From what I can tell, there is no technical conflict either way — this is a question of operational preference.
```
deven96: Persistence would be replaced (either in cluster mode or not) by the cluster snapshot mechanism right?

docs/specs/replication.md
```diff
@@ -0,0 +1,859 @@
...
+6. Once caught up, the node is automatically promoted to a voter.
+
+```bash
+# Add a 4th node to an existing cluster
+ahnlich-db run --port 1372 --cluster-addr 127.0.0.1:9004 --cluster-join 127.0.0.1:9001
+```
+
+### Graceful Shutdown
+
+When a node receives SIGTERM:
```
deven96: Not just a sigterm, we may also be panicking internally... I think @Iamdavidonuh introduced a way to cleanup during panics/shutdown via the Task mechanism in https://github.com/deven96/ahnlich/pull/323/changes

docs/specs/replication.md
```diff
@@ -0,0 +1,859 @@
...
+                          │    → Log entry committed and applied on all nodes
+                          │
+                          └─ Return leader's response to client
+```
+
+This transparent forwarding means clients can connect to any node in the cluster without needing to discover the leader. The trade-off is an extra network hop for writes that hit followers.
+
+### Pipeline Operations
+
+The `Pipeline` RPC accepts a sequence of operations and executes them sequentially. In cluster mode:
```
jimezesinachi, line 342: Pipeline handling

docs/specs/replication.md
jimezesinachi: Done

docs/specs/replication.md
jimezesinachi: > I think Pipeline calls should also be forwarded to the master node... as there's no way to tell on the protocol level if the pipeline contains a mutation or not

This is already the case, see here: https://github.com/deven96/ahnlich/pull/325/changes#r3047735725

docs/specs/replication.md
jimezesinachi: Eventually, yes (so that we ensure we are not breaking userspace first before making the replacement), the question is how do we manage it in the meanwhile

deven96: Given that both persistence methods (i.e our existing one and the new RAFT one) should end up with the same underlying representation of the store as serialised by serde_json. I'd say it shouldn't be so difficult to replace then no?

jimezesinachi: > Given that both persistence methods (i.e our existing one and the new RAFT one) should end up with the same underlying representation of the store as serialised by serde_json. I'd say it shouldn't be so difficult to replace then no?

Sure, but the question is whether we should keep it enabled in cluster mode or not?

deven96: So the way I see it is

- Cluster mode we use the provided abstraction which underneath dumps the state to serde_json or whatever we chose
- Non cluster mode with persistence ...underlying persistence still uses the same cluster mode persistence function 

This way whenever we think about persistence, we only have to think about serialization/deserialization and then affect said function

jimezesinachi: So you're saying we should not write/read from `utils::Persistence` at all in cluster mode, yes?

deven96: Correct... and in non cluster mode we should (but reusing the same nubbins)

jimezesinachi: Understood

Iamdavidonuh (approved):

PR #332, opened 2026-05-08
feat: Milestone 1 - added replication plumbing, tests and shared helpers

jimezesinachi:


ahnlich/replication/src/types.rs
```diff
@@ -0,0 +1,65 @@
...
+pub enum DbCommand {
+    CreateStore(Vec<u8>),
+    CreatePredIndex(Vec<u8>),
+    CreateNonLinearAlgorithmIndex(Vec<u8>),
+    Set(Vec<u8>),
+    DelKey(Vec<u8>),
+    DelPred(Vec<u8>),
+    DropPredIndex(Vec<u8>),
+    DropNonLinearAlgorithmIndex(Vec<u8>),
+    DropStore(Vec<u8>),
```
deven96, line 34: I thought we added PurgeStores recently to DB?

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -0,0 +1,425 @@
+// `clippy::result_large_err` and `clippy::type_complexity` fire on signatures
```
deven96, line 1: Prefix mod with a definition of what the statemachine is and what it does

ahnlich/replication/src/types.rs
jimezesinachi: A'firm, good catch

jimezesinachi: Turns out, there appears to be no purge_stores handler for DB. There is one for AI, but none for DB. See here: https://github.com/deven96/ahnlich/blob/main/ahnlich/db/src/server/handler.rs

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    fn apply_to_state_machine_sync(
+        &mut self,
+        entries: &[C::Entry],
+    ) -> Result<Vec<C::R>, StorageError<C::NodeId>> {
+        let mut inner = self.inner.lock().expect("state machine lock poisoned");
+        let mut responses = Vec::new();
+
+        for entry in entries {
```
deven96, line 254: A bit skeptical about running the entries in a loop to inner.handler.apply. If there's an error with applying entry in the middle, we return error but have already applied some previous entries?

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -0,0 +1,447 @@
...
+        meta: &SnapshotMeta<C::NodeId, C::Node>,
+        snapshot: Box<C::SnapshotData>,
+    ) -> Result<(), StorageError<C::NodeId>> {
+        let cursor: Cursor<Vec<u8>> = (*snapshot).clone().into();
+        let data = cursor.into_inner();
+        let decoded = deserialize_snapshot::<H::Snapshot>(&data).map_err(|e| StorageError::IO {
+            source: StorageIOError::read_state_machine(&e),
+        })?;
+
+        let mut inner = self.inner.lock().expect("state machine lock poisoned");
```
deven96: Curious about panicking as a whole with respect to lock poisoning of the state machine

Shouldn't it be captured with `StorageError` instead?

ahnlich/replication/src/types.rs
deven96: Oh schade... we definitely don't have then
Skip for now

ahnlich/replication/src/storage/rocksdb.rs
```diff
@@ -0,0 +1,382 @@
...
+    _p: std::marker::PhantomData<C>,
+}
+
+impl<C: RaftTypeConfig> RocksLogStore<C> {
+    pub fn open(path: impl AsRef<Path>) -> Result<Self, StorageError<C::NodeId>> {
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
+        opts.create_missing_column_families(true);
+        let cfs = vec![
+            ColumnFamilyDescriptor::new("logs", Options::default()),
```
deven96: Would rather we have static strings referencing the column names and reuse across the rocksdb mod

ahnlich/replication/src/storage/rocksdb.rs
```diff
@@ -0,0 +1,382 @@
...
+            None
+        };
+        let purged = purged.unwrap_or(None);
+        Ok(LogState {
+            last_purged_log_id: purged.clone(),
+            last_log_id: last.or(purged),
+        })
+    }
+
+    fn save_vote_sync(&mut self, vote: &Vote<C::NodeId>) -> Result<(), StorageError<C::NodeId>> {
```
deven96: Perhaps a stronger harness around setting&getting meta.... I'm largely wary about having to repeat strings as there's a huge chance that with an update things can be missed

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -0,0 +1,447 @@
...
+    ) -> Result<Box<C::SnapshotData>, StorageError<C::NodeId>> {
+        Ok(Box::new(C::SnapshotData::from(Cursor::new(Vec::new()))))
+    }
+
+    fn install_snapshot_sync(
+        &mut self,
+        meta: &SnapshotMeta<C::NodeId, C::Node>,
+        snapshot: Box<C::SnapshotData>,
+    ) -> Result<(), StorageError<C::NodeId>> {
+        let cursor: Cursor<Vec<u8>> = (*snapshot).clone().into();
```
deven96: Do we have to clone the entire snapshot in order to deserialize it?

ahnlich/replication/src/storage/state_machine.rs
jimezesinachi: > A bit skeptical about running the entries in a loop to inner.handler.apply. If there's an error with applying entry in the middle, we return error but have already applied some previous entries?

Shey, I see what you're saying. I'm thinking of potential solutions: maybe rollback, or just try and check and fail early somehow?

ahnlich/replication/src/storage/state_machine.rs
jimezesinachi: > Curious about panicking as a whole with respect to lock poisoning of the state machine
> 
> Shouldn't it be captured with `StorageError` instead?

I see what you're saying. I think my thinking here was that if lock is poisoned, the state machine is corrupt and in a way "unrecoverable until node restart". But user API wise, it does make sense to just bubble up as a `StorageError`, possibly with a message saying "the state machine lock is poisoned"

ahnlich/replication/src/storage/rocksdb.rs
jimezesinachi: Makes sense. I'll use constants

ahnlich/replication/src/storage/rocksdb.rs
jimezesinachi: Makes sense. I'll use helper functions here

ahnlich/replication/src/storage/state_machine.rs
jimezesinachi: Looking again, I'm not sure. Will take a closer look and fix

ahnlich/replication/src/storage/state_machine.rs
deven96: Yeah maybe fail early for a start as I'm not sure we have some sure fire way to rollback yet

jimezesinachi: > Yeah maybe fail early for a start as I'm not sure we have some sure fire way to rollback yet

I agree on the rollback, however on closer look, I don't think we can really fail any earlier in this scope. At this point, even if `handler.apply()` fails, the log is already committed to storage, so any failure that happens would be likely due to the things like the command/data itself being invalid. The "fail early" will have to happen in the AI/DB handler, which we haven't added yet (not in this PR). I think the only thing we can do here is the check for state machine lock, which I've now added

deven96 (approved):

deven96:
@Iamdavidonuh please one more pass

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -0,0 +1,485 @@
...
+    // `StateMachineStore` replays committed entries sequentially, not
+    // transactionally across a batch. Earlier entries in the same batch may
+    // already be reflected in application state if a later one fails.
+    //
+    // Because of that, ordinary domain/precondition validation is expected to
+    // happen before replication. Errors returned from `apply()` should
+    // therefore represent invariant violations, corruption, or
+    // storage/infrastructure failures rather than routine business-rule
+    // rejection.
+    fn apply(&mut self, data: &C::D) -> Result<C::R, StorageError<C::NodeId>>;
```
Iamdavidonuh, line 84: Per your point in https://github.com/deven96/ahnlich/pull/332#discussion_r3257523417, if validation happens pre-`client_write` in M2, `apply()` should never see an invalid command by construction. Which raises the question of what the trait should return.

If `apply()` returns `Err`, openraft's state machine worker exits permanently. No retry; apply pipeline stays dead until restart.

`openraft-0.9.24/src/core/sm/worker.rs:57-71`:
```rust
fn do_spawn(mut self) -> JoinHandleOf<C, ()> {
    C::AsyncRuntime::spawn(async move {
        let res = self.worker_loop().await;
        if let Err(err) = res {
            tracing::error!("{} while execute state machine command", err);
            let _ = self.resp_tx.send(Notify::StateMachine {
                command_result: CommandResult { command_seq: 0, result: Err(err) },
            });
        }
    })
}
```

Two shapes worth weighing:

1. Drop the `Result` and have `C::R` itself encode per-command rejection. Contract is explicit at the type level: applying a committed entry never fails.
2. Keep `Result` only for genuine storage/invariant failures, and pin the rule in the trait doc that handler validation must happen pre-`client_write`. In that case the `failed_apply_does_not_advance_last_applied` test should go, since it asserts a behaviour that would kill the worker in practice.

ahnlich/replication/src/network.rs
```diff
@@ -0,0 +1,201 @@
...
+impl<C: RaftTypeConfig> RaftNetwork<C> for GrpcRaftNetwork<C> {
+    async fn append_entries(
+        &mut self,
+        rpc: AppendEntriesRequest<C>,
+        _option: RPCOption,
+    ) -> Result<AppendEntriesResponse<C::NodeId>, RPCError<C::NodeId, C::Node, RaftError<C::NodeId>>>
+    {
+        let mut client = self.client.clone();
+        let payload = PbAppendEntriesRequest {
+            payload: to_vec(&rpc).map_err(rpc_err::<C, _>)?,
```
Iamdavidonuh: Question: do we want JSON on the Raft RPC hot path?

For snapshots, JSON matches `utils::persistence`, so that's consistent. For AppendEntries it's different. From M2, `C::D` will be `DbCommand::Set(Vec<u8>)`, and `serde_json` encodes `Vec<u8>` as a JSON array of integers. I checked empirically:

```rust
let bytes: Vec<u8> = vec![72, 101, 108, 108, 111, 0, 255, 1, 2, 3];
let json = serde_json::to_vec(&bytes).unwrap();
// json as text: [72,101,108,108,111,0,255,1,2,3]
// 10 input bytes -> 32 json bytes

let big: Vec<u8> = (0..=255u8).cycle().take(1024).collect();
let big_json = serde_json::to_vec(&big).unwrap();
// 1024 bytes -> 3657 json bytes (ratio 3.57x)
```

About 3.5x expansion, paid on every AppendEntries in both directions.

Bitcode is already in the repo (PR #321), so swapping the codec here is small. Once M2 ships and real traffic flows, changing it becomes a rolling-upgrade break.

Worth doing now while there are no consumers, or defer to M2 when we can measure?

ahnlich/replication/src/node.rs
```diff
@@ -0,0 +1,34 @@
...
+impl From<&NodeInfo> for ReplicationNode {
+    fn from(value: &NodeInfo) -> Self {
+        Self {
+            raft_addr: value.raft_addr.clone(),
+            service_addr: value.service_addr.clone(),
+        }
+    }
+}
+
+impl From<ReplicationNode> for BasicNode {
```
Iamdavidonuh: Do we plan to use `BasicNode` anywhere? Tests and admin use `ReplicationNode` directly. If we're not using it, we can drop the impl, especially since `BasicNode` has no service address field, so `ClusterInfo` would surface an empty one if it ever became `C::Node`.

ahnlich/replication/src/network.rs
jimezesinachi: I agree, the original version used bincode, but Deven pointed out that the package was no longer maintained, and we discussed using postcard, but we decided to hold it until we could confirm it was better than serde_json. I didn't even realise we had another solution (bitcode) in the repo already, 3.57x is definitely significant, so we should look to swap usage early now, I think

Iamdavidonuh (approved):

PR #335, opened 2026-05-18
feat(ai): add MIGraphX execution provider for AMD GPUs

Benehiko:
> **Note:** This PR was generated with Claude. It has been compile-checked, unit-tested, regenerated for the Go/Node/Python SDKs, and **validated end-to-end on AMD hardware** (Radeon RX 9070, gfx1201/RDNA4, ROCm 6.4.2) using the `Dockerfile.rocm` image included in this PR.

## Summary

Adds an opt-in execution provider so ahnlich-ai can target AMD GPUs:

- `MIGRAPHX` — AMD's supported execution provider for `onnxruntime >= 1.23` ([MIGraphX EP docs](https://onnxruntime.ai/docs/execution-providers/MIGraphX-ExecutionProvider.html))

It wires through the same path the existing `CUDA`, `TENSOR_RT`, `DIRECT_ML`, and `CORE_ML` providers use. Provider registration is compiled in via a new opt-in cargo feature (`migraphx`) that mirrors the existing `tensorrt`/`cuda`/`coreml`/`directml` feature pattern — default builds are unchanged.

> **History:** earlier revisions of this PR also added a `ROCM` provider. Upstream onnxruntime removed the ROCm execution provider in 1.23 in favor of MIGraphX, and the maintainers plan to bump the `ort` pin past that release, so `ROCM` was dropped from this PR before merge (proto tag `4` and the name `ROCM` are `reserved`). The ORT pin bump itself is intentionally **out of scope** here and will land separately.

### Protocol + Rust core

- `protos/ai/execution_provider.proto` — add `MIGRAPHX = 5`; `reserved 4` / `reserved "ROCM"` so the retired tag can never be reused
- `protos/README.md` — mention `MIGRAPHX` alongside `CUDA` / `TENSOR_RT`
- `ahnlich/types/src/ai/execution_provider.rs` — regenerated Rust enum
- `ahnlich/ai/src/engine/ai/providers/ort/mod.rs` — register `MIGraphXExecutionProvider` through `InnerAIExecutionProvider::MIGraphX`
- `ahnlich/ai/Cargo.toml` — two new opt-in features:
  - `migraphx = ["ort/migraphx"]` — compile in provider registration (same pattern as the existing EP features)
  - `ort-load-dynamic = ["ort/load-dynamic"]` — load `libonnxruntime` via `dlopen` (`ORT_DYLIB_PATH`) instead of linking at build time. Needed when the runtime library is a different ORT release than the one `ort` was built against (e.g. AMD's builds): link-time ELF symbol versioning (`VERS_1.19.0`) rejects the version skew that the forward-compatible ORT C API happily serves.

### DSL grammar + parser + tests

- `ahnlich/dsl/src/syntax/syntax.pest` — extend the `execution_provider` rule so the pest tokeniser emits `migraphx`
- `ahnlich/dsl/src/ai.rs` — accept `"migraphx"` in `parse_to_execution_provider`
- `ahnlich/dsl/src/tests/ai.rs` — round-trip test mirroring the existing TensorRT / CUDA tests

### Packaging

- `ahnlich/Dockerfile.rocm` — AMD sibling of the default (CUDA-flavour) `Dockerfile`. Runtime stage is `rocm/dev-ubuntu-24.04:6.4.2` (6.4.2 is the first ROCm release with RDNA4/gfx1201 support) carrying:
  - the shared libraries from AMD's prebuilt `onnxruntime_rocm` 1.21.0 wheel (`repo.radeon.com`, `rocm-rel-6.4.2`)
  - the `hipblas`/`hipfft`/`hipsparse`/`hiprand`/`rocrand`/`rccl`/`migraphx` runtime packages that the provider libraries `dlopen` at session creation (each verified empirically)
  - `ahnlich-ai` built with `--features migraphx,ort-load-dynamic` and `ORT_DYLIB_PATH` pointed at the AMD library

### SDK regeneration

- `sdk/ahnlich-client-go/grpc/ai/execution_provider/execution_provider.pb.go` — regenerated via `buf generate`
- `sdk/ahnlich-client-node/grpc/ai/execution_provider_pb.ts` — regenerated via `buf generate`
- `sdk/ahnlich-client-py/ahnlich_client_py/grpc/ai/execution_provider/__init__.py` — regenerated (betterproto), including the proto doc comments as docstrings

### Docs

- `README.md` — document MIGraphX prerequisites under "Execution Providers", including the note that upstream ORT removed the older ROCm provider in 1.23

## Verification

- `cargo test -p dsl` — **34 / 34 passing** (round-trips `"migraphx"` end-to-end through `parse_ai_query`)
- `cargo fmt` / `cargo clippy` — clean, with default features and with `migraphx,ort-load-dynamic`
- `cargo build -p ahnlich_types` — no diff beyond the intended enum change (generated code in sync with the proto)
- **End-to-end on AMD hardware** (Radeon RX 9070, gfx1201/RDNA4, ROCm 6.4.2, `Dockerfile.rocm` image against a stock `ahnlich-db` container, re-run 2026-08-14 after the ROCm removal):
  - `CREATESTORE` with `all-minilm-l6-v2`, `SET` with `executionprovider migraphx` (3 entries embedded on GPU), then `GETSIMN` returns the correct top result with sensible similarity ordering
  - GPU usage confirmed via `rocm-smi --showpids` (KFD process holding ≈ 0.6 GiB VRAM) — no silent CPU fallback

## Open questions for maintainers

- Should CI build/publish the `Dockerfile.rocm` image (release matrix entry)? The runtime stage is ~13 GB unpacked (ROCm device kernels dominate), which may warrant an on-demand rather than per-release build.
- Validation above is RDNA4; a sanity run on Instinct (CDNA) hardware would be nice-to-have but the wheel targets those architectures first-class.

## References

- ONNX Runtime MIGraphX Execution Provider docs: <https://onnxruntime.ai/docs/execution-providers/MIGraphX-ExecutionProvider.html>
- ONNX Runtime ROCm EP docs (removal in 1.23): <https://onnxruntime.ai/docs/execution-providers/ROCm-ExecutionProvider.html>
- `ort` crate `MIGraphXExecutionProvider`: <https://docs.rs/ort/2.0.0-rc.5/ort/struct.MIGraphXExecutionProvider.html>
- AMD ROCm compatibility matrix: <https://rocm.docs.amd.com/en/latest/compatibility/compatibility-matrix.html>
- AMD prebuilt ORT wheels: <https://repo.radeon.com/rocm/manylinux/rocm-rel-6.4.2/>

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Iamdavidonuh:
This looks interesting.

CC: @Ayobami-00

deven96:
Thanks for this! @Benehiko 

> Keep both ROCM and MIGRAPHX, or drop ROCM when the ORT pin moves past 1.23?

I think we should drop ROCM when ORT pin moves past 1.23

> Should CI build/publish the Dockerfile.rocm image (release matrix entry)? The runtime stage is ~13 GB unpacked (ROCm device kernels dominate), which may warrant an on-demand rather than per-release build

If we get rid of ROCM does that drop the runtime stage? I think per-release builds would be appreciated

> Validation above is RDNA4; a sanity run on Instinct (CDNA) hardware would be nice-to-have but the wheel targets those architectures first-class.

I think the sanity check is good enough for this iteration... unless anything comes to mind which may cause it not to run on CDNA

jimezesinachi:
Go and python tests are failing @Benehiko

deven96:
> Go and python tests are failing @Benehiko

Lints i think

jimezesinachi (approved):
LGTM

deven96:
Perhaps we should move the ORT pin on this PR so we can get rid of ROCM as well and also add the CI for per-release builds .... what do you think @jimezesinachi @Benehiko

jimezesinachi:
> Perhaps we should move the ORT pin on this PR so we can get rid of ROCM as well and also add the CI for per-release builds .... what do you think @jimezesinachi @Benehiko

Sounds good to me

Benehiko:
I've dropped the ROCM dependency and kept it to MIGraphX instead. This should unpin the ORT. A follow-up can update the ORT library.

PR #336, opened 2026-05-20
feat(charts): Helm charts for ahnlich-db, ahnlich-ai, and umbrella

### LINKED ISSUE #272: Native k8s support for ahnlich
Perhaps investing time into creating a helmchart?

Iamdavidonuh:
Closes #272.

## Summary

Adds three Helm charts under `charts/` for deploying ahnlich on Kubernetes:

- **`charts/ahnlich-db/`** — StatefulSet, Service, optional headless Service + PodDisruptionBudget for cluster mode, optional Ingress, optional Gateway API `GRPCRoute`. Supports standalone (one replica + JSON snapshot persistence) and Raft cluster modes via a single `cluster.enabled` toggle.
- **`charts/ahnlich-ai/`** — same shape as DB, plus required `db.host` wiring, a dedicated PVC for the model cache, an `initContainer` that blocks startup until the DB Service is reachable (matching docker-compose's `depends_on: service_healthy`), and `--agent ai` on probes.
- **`charts/ahnlich/`** — umbrella that aggregates the two sub-charts and pre-wires `ahnlich-ai.db.host` to the DB Service when installed with release name `ahnlich`. Optionally deploys an in-cluster OTLP tracing backend (Jaeger all-in-one by default, image-swappable for any OTLP receiver) for the demo/local path. Template-time `fail` guards the pinned-release-name footgun with an explicit override command.

## Architecture decisions

- **Pattern A values forwarding** (explicit per-sub-chart blocks like `--set ahnlich-db.cluster.enabled=true`, no `global:` magic). Verbose but transparent; allows mixed configurations (DB via Gateway, AI via Ingress, or any combination).
- **No bundled Ingress Controller or Gateway implementation.** Charts ship `Ingress` and `GRPCRoute` resource templates only; controllers are cluster-wide infrastructure managed by platform teams.
- **M2-ready cluster mode.** The chart renders correct YAML when `cluster.enabled=true`. Once the ahnlich binary wires the `--cluster-*` flags (replication milestone M2), the same chart deploys without changes. Until then, enabling cluster mode crashes the pod at the binary level with `unknown argument '--cluster-addr'` — verified.
- **Logging via `--log-level` CLI flag, not `RUST_LOG`.** Chart exposes `logLevel` value that maps to the flag (ahnlich doesn't read `RUST_LOG`). `env`/`envFrom` blocks are separate for other process env.
- **Image pull policy auto-switches to `Always` for `:latest`** so `helm upgrade` actually rolls. Tagged versions use `IfNotPresent`.

## What was verified live on Rancher Desktop

| Scenario | Result |
|---|---|
| `helm install charts/ahnlich-db` standalone | DB up, PVC bound, PING via `--agent db` succeeds |
| `service.type=LoadBalancer` external access | External IP assigned, no port-forward needed |
| Tracing enabled + in-cluster Jaeger | Traces visible in Jaeger UI |
| `helm install charts/ahnlich-ai` pointed at DB | AI up, CREATESTORE flows AI → DB, store visible from both sides |
| `helm install charts/ahnlich` umbrella, standalone | DB + AI up, initContainer waits for DB, zero AI restarts |
| Gateway API path via Envoy Gateway + GRPCRoute | gRPC PING through `192.168.64.4:8080` → PONG |
| `helm install charts/ahnlich` cluster mode | Pod crashes at the binary as expected (M1 limitation, documented) |
| Release name mismatch + missing override | `helm install` fails at template time with override command |

## Notes

- Sub-chart dependencies use local `file://` paths; run `helm dependency build charts/ahnlich` after cloning before installing.
- README in each chart directory covers values reference, external access patterns, and operational caveats (TLS recommendation, VCT immutability with persistence toggle, sub-chart version bump workflow).

### DIFF
--- a/charts/ahnlich-ai/README.md
+++ b/charts/ahnlich-ai/README.md
@@ -1,0 +1,181 @@
+# ahnlich-ai
+
+Helm chart for `ahnlich-ai`, the embedding / AI proxy that sits in front of `ahnlich-db`.
+
+Architecturally, `ahnlich-ai` accepts raw inputs (text / image / audio), turns them into embeddings via ONNX models, and forwards everything to a backing `ahnlich-db` cluster over gRPC. This chart deploys the AI process; it does **not** deploy `ahnlich-db` — point `db.host` at an existing DB.
+
+The chart supports two deployment modes:
+
+- **Standalone** (default): single replica with periodic JSON-snapshot persistence of AI-local store metadata.
+- **Cluster (Raft)**: N replicas forming a Raft cluster for AI-local state. Activate with `--set cluster.enabled=true`.
+
+## Prerequisites
+
+- Kubernetes 1.20+ (for `appProtocol: grpc`).
+- A default `StorageClass` (the chart falls back to it unless overridden).
+- A reachable `ahnlich-db` deployment. Point `db.host` at its Service DNS name.
+- Optional, for external L7 access: an Ingress Controller or Gateway API implementation.
+
+## Quick start
+
+```bash
+# Standalone AI talking to an existing DB
+helm install my-ai charts/ahnlich-ai \
+  --set db.host=my-ahnlich-ahnlich-db
+
+# With tracing
+helm install my-ai charts/ahnlich-ai \
+  --set db.host=my-ahnlich-ahnlich-db \
+  --set tracing.enabled=true \
+  --set tracing.otelEndpoint=http://jaeger-collector:4317
+
+# Exposed externally
+helm install my-ai charts/ahnlich-ai \
+  --set db.host=my-ahnlich-ahnlich-db \
+  --set service.type=LoadBalancer
+```
+
+## Values
+
+### Image
+
+| Key | Default | Description |
+|---|---|---|
+| `image.repository` | `ghcr.io/deven96/ahnlich-ai` | Image repository. |
+| `image.tag` | `latest` | Image tag. |
+| `image.pullPolicy` | `IfNotPresent` | Standard k8s pull policy. |
+
+### Backing DB
+
+| Key | Default | Description |
+|---|---|---|
+| `db.host` | `""` | **Required.** DNS name of the backing ahnlich-db Service. |
+| `db.port` | `1369` | DB port. |
+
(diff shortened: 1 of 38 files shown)


charts/ahnlich-ai/templates/statefulset.yaml
```diff
@@ -0,0 +1,195 @@
...
+    {{- include "ahnlich-ai.labels" . | nindent 4 }}
+spec:
+  serviceName: {{ include "ahnlich-ai.fullname" . }}{{ if .Values.cluster.enabled }}-headless{{ end }}
+  replicas: {{ if .Values.cluster.enabled }}{{ .Values.cluster.replicas }}{{ else }}1{{ end }}
+  selector:
+    matchLabels:
+      {{- include "ahnlich-ai.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
```
MeNsaaH, line 15: need to add a value for additional Pod labels. something like `.Values.additionalPodLabels` or `.Values.podLabels` can suffice

charts/ahnlich-ai/README.md
```diff
@@ -0,0 +1,185 @@
...
+  --set tracing.enabled=true \
+  --set tracing.otelEndpoint=http://jaeger-collector:4317
+
+# Exposed externally
+helm install my-ai charts/ahnlich-ai \
+  --set db.host=my-ahnlich-ahnlich-db \
+  --set service.type=LoadBalancer
+```
+
+## Values
```
MeNsaaH, line 38: preferably use https://github.com/norwoodj/helm-docs to generate this section

MeNsaaH (commented):
Looks good overall. Just a few 💅🏿 tweaks

charts/ahnlich-ai/templates/statefulset.yaml
Iamdavidonuh: done!

charts/ahnlich-ai/README.md
Iamdavidonuh: done!

deven96 (approved):

PR #338, opened 2026-06-01
feat(db): add db cluster logic and tests

jimezesinachi:


ahnlich/db/src/server/handler.rs
```diff
@@ -36,6 +46,7 @@ pub struct Server {
     client_handler: Arc<ClientHandler>,
     task_manager: Arc<TaskManager>,
     config: ServerConfig,
+    cluster: Option<ClusterRuntime>,
```
deven96: ClusterRuntime would ideally always contain store_handler.... and we typically only ever want to access store_handler directly independent of cluster if cluster doesn't exist? Seems like we should have a cleaner 

```rust

enum StoreRuntime {
    Standalone(StoreHandler), // perhaps a more befitting rename to handler/ runtime to be consistent
    Cluster(ClusterRuntime), 
}
```

ahnlich/db/src/server/cluster_queries.rs
```diff
@@ -0,0 +1,118 @@
...
+            .with_handler(|handler| f(handler.store_handler()))
+            .map_err(|err| map_storage_error("failed to access clustered state machine", err))?
+            .map_err(Into::into)
+    } else {
+        f(store_handler).map_err(Into::into)
+    }
+}
+
+pub(crate) async fn list_stores_response(
+    cluster: Option<&ClusterRuntime>,
```
deven96: Not particularly sold on having optional cluster and a reference to store_handler when cluster always contains handler... see above comment

ahnlich/utils/src/server.rs
```diff
@@ -55,6 +55,10 @@ pub trait AhnlichServerUtils: BlockingTask + Sized + Send + Sync + 'static + Deb
 
     fn task_manager(&self) -> Arc<TaskManager>;
 
+    fn should_spawn_persistence(&self) -> bool {
```
deven96, line 58: FIXME on this so we don't forget while implementing AI db cluster logic

ahnlich/db/src/tests/cluster_tests.rs
```diff
@@ -0,0 +1,319 @@
...
+    assert_eq!(
+        cluster_info
+            .nodes
+            .iter()
+            .filter(|node| node.role == NodeRole::Leader as i32)
+            .count(),
+        1
+    );
+
+    let list_stores_error = follower_one_client
```
deven96, line 323: Leader client shouldn't error right? Perhaps we assert that here as well

deven96 (commented):
Some minor comments

ahnlich/db/src/server/handler.rs
jimezesinachi: Actually agree with this, I went in the direction of not changing the shape of Server, but this is the better direction long term

ahnlich/db/src/server/cluster_queries.rs
jimezesinachi: A'firm, I will change the shape now

ahnlich/db/src/tests/cluster_tests.rs
jimezesinachi: Make sense, adding now

ahnlich/db/src/server/handler.rs
```diff
@@ -654,19 +797,28 @@ impl AhnlichServerUtils for Server {
         self.task_manager.clone()
     }
 
+    fn should_spawn_persistence(&self) -> bool {
```
jimezesinachi, line 798: See here

ahnlich/db/src/replication/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        ))),
+    }
+}
+
+impl StateMachineHandler<DbTypeConfig> for DbStateMachine {
+    type Snapshot = Stores;
+
+    fn apply(&mut self, data: &DbCommand) -> Result<DbResponse, StorageError<u64>> {
+        match data {
+            DbCommand::CreateStore(payload) => {
```
deven96, line 98: There's probably a nice way to turn decode_payload into a macro so we dont have to manually pass that string but can extract it from the ty via stringify!

deven96: Something like

```rust
macro_rules! decode_payload {
    ($ty:ty, $payload:expr) => {{
        <$ty>::decode($payload).map_err(|err| StorageError::IO {
            source: StorageIOError::read_state_machine(&std::io::Error::other(format!(
                "failed to decode {} raft payload: {}",
                stringify!($ty),
                err,
            ))),
        })
    }};
}

let params: query::CreateStore =
    decode_payload!(query::CreateStore, payload)?;
```

ahnlich/db/src/server/cluster_mutations.rs
```diff
@@ -0,0 +1,65 @@
+use crate::server::cluster::ClusterRuntime;
+use ahnlich_replication::node::ReplicationNode;
+use ahnlich_replication::types::{DbCommand, DbResponse};
+use openraft::error::RaftError;
+use serde::de::DeserializeOwned;
+use std::any::{Any, TypeId};
+
+pub(crate) fn map_client_write_error(
+    command_name: &str,
```
deven96, line 25: Another good use of macros to enforce not needing command name alongside the actual command

ahnlich/db/Cargo.toml
```diff
@@ -42,6 +42,10 @@ pulp.workspace = true
 papaya.workspace = true
 ahnlich_types = { path = "../types", version = "*"  }
 tonic.workspace = true
+prost.workspace = true
+openraft.workspace = true
+ahnlich-replication = { path = "../replication", version = "*" }
+bitcode = { version = "0.6", features = ["serde"] }
```
deven96: We should move to general Cargo.toml... my suspicion is we will reuse elsewhere

ahnlich/db/src/replication/mod.rs
jimezesinachi: Makes sense, will implement this

PR #339, opened 2026-06-03
fix: wrap papaya collection constructors in catch_unwind to prevent silent crashes

### LINKED ISSUE #100: Fallible collections

gregorian-09:
# fix: wrap papaya collection constructors in catch_unwind to prevent silent crashes

Closes #318
Closes #100 (papaya fallible collections portion)

## Problem

Papaya's `HashMap::new()` and `HashSet::new()` are lazy. They allocate an internal hash table of capacity 0 at construction time and defer the real allocation to the first `insert()`. The first insert triggers `Table::alloc()`, which contains `assert!(len.is_power_of_two())`. If this assertion fails, or if `seize::Collector::new()` (called internally by papaya on first use) panics due to thread creation failure or OOM,
the crash happens at an arbitrary mutation point far from construction.

This is especially dangerous because the panic hook that would catch and log these failures is installed in `Server::start()`, which runs **after** `StoreHandler::new()` and `AIStoreHandler::new()`. Any papaya initialization panic in the constructors escapes without the hook, either aborting the process with an inscrutable message or unwinding through
foreign code (C ABI entry points), causing undefined behavior.

### Core approach

Added a new `utils::fallible` module (`ahnlich/utils/src/fallible.rs`) with four helpers that force immediate allocation and catch panics:

- `try_new_hashmap()` -- constructs `ConcurrentHashMap::with_capacity(1)` inside `std::panic::catch_unwind`. With `capacity=1`, papaya calls `entries_for(1)` which returns `next_power_of_two(1) = 2`, a valid power of two, triggering actual table allocation at construction time.
- `try_new_arc_hashmap()` -- convenience wrapper returning
  `Arc<ConcurrentHashMap>`.
- `try_new_hashset()` -- same pattern for `ConcurrentHashSet`.
- `try_new_hashset_with_capacity(cap)` -- variant that lets callers specify the initial capacity (used by `ClientHandler`).

All four return `Result<T, String>`.

### Design decision: no public API break

My first approach changed `StoreHandler::new()` and
`AIStoreHandler::new()` to return `Result<Self, String>`. This was cleaner (the error propagated to `main()` for a user-friendly message) but it broke the public API. Any downstream crate or binary using these constructors would fail to compile.

I changed tack: the constructors keep their original `-> Self` signature. Inside, they call `try_new_*().unwrap_or_else(|e| {
eprintln!("Fatal: ..."); std::process::abort(); })`. This preserves full backward compatibility while guaranteeing that a papaya init failure prints a clear diagnostic and immediately stops the process (which is the only safe thing to do when your core data structures can't be allocated).

For constructors that run **after** the panic hook is installed
(`Store::create()`, `NonLinearAlgorithmIndices::create()`,
`PredicateIndices::init()`, `PredicateIndex::init()`, `HNSW`
constructors), I used `try_new_*().expect("...")`. If papaya init fails there, the panic hook catches the re-panic and handles it gracefully (logs, cleans up tasks, etc.).

### Wrapped papaya sites

| Location | Constructor | Running before panic hook? | Handling |
|---|---|---|---|
| `db/src/engine/store.rs:StoreHandler::new()` | `try_new_arc_hashmap` | Yes | `abort()` with message |
| `db/src/engine/store.rs:Store::create()` | `try_new_hashmap` | No (inside task) | `expect()` |
| `db/src/algorithm/non_linear.rs:NonLinearAlgorithmIndices::create()` | `try_new_hashmap` | No | `expect()` |
| `db/src/engine/predicate.rs:PredicateIndices::init()` | `try_new_hashmap` + `try_new_hashset` | No | `expect()` |
| `db/src/engine/predicate.rs:PredicateIndex::init()` | `try_new_hashmap` | No | `expect()` |
| `db/src/engine/predicate.rs:PredicateIndex::add()` | `try_new_hashset` | No (inside rayon) | `expect()` |
| `ai/src/engine/store.rs:AIStoreHandler::new()` | `try_new_hashmap` | Yes | `abort()` with message |
| `utils/src/client.rs:ClientHandler::new()` | `try_new_hashset_with_capacity` | Yes | `abort()` with message |
| `similarity/src/hnsw/index.rs:7 call sites` | Private `new_hashmap()` helper | No (runtime) | `expect()` |

## Tests added

26 tests in `utils::fallible::tests`:

**Happy path**
- `hashmap_put_get` -- put and get a key-value pair
- `hashmap_empty` -- freshly constructed map is empty
- `hashset_insert_contains` -- insert and check membership

**Type variants**
- `hashmap_string_keys` -- string keys (primary use case)
- `hashmap_complex_key_type` -- custom `Hash + Eq` struct keys

**Capacity edge cases**
- `hashset_with_capacity_zero` -- capacity 0 (papaya does not preallocate)
- `hashset_with_capacity_one` -- capacity 1 (minimum to force allocation)
- `hashset_with_capacity_large` -- capacity 10,000
- `hashset_with_capacity_zero_still_holds_data` -- 100 inserts into zero-capacity set

**Arc variant**
- `arc_hashmap_put_get` -- `try_new_arc_hashmap` round-trip
- `arc_hashmap_is_shareable` -- Arc can be immutably borrowed

**Concurrency**
- `concurrent_construction` -- 16 threads each create their own map
- `concurrent_hashset_construction` -- 16 threads each create their own set
- `concurrent_writes_to_shared_map` -- 10 threads insert into a shared map, verify all 10 entries visible
- `concurrent_reads_and_writes` -- 4 writers + 4 readers on shared map
- `hashset_concurrent_stress` -- 8 threads, 1000 total inserts

**catch_unwind integration**
- `catch_unwind_wrapper_catches_panics` -- verify catch_unwind correctly returns `Err` on a manual panic
- `unwind_safe_compatibility` -- papaya operations inside
  `AssertUnwindSafe` closure do not panic
- `fallible_catches_papaya_init_panic` -- verify the error-propagation pattern works end to end

**Lifecycle**
- `repeated_construction` -- 100 sequential create/drop cycles of maps
- `repeated_construction_hashset` -- same for sets (catches seize collector state leaks across init/teardown)

**API contract**
- `hashmap_returns_ok` -- `try_new_hashmap` returns `Ok` when
  allocation succeeds
- `hashset_returns_ok` -- same for set
- `hashset_with_capacity_returns_ok` -- same for with_capacity variant

**Iteration and reserve**
- `hashmap_iteration` -- iterating a map built via fallible constructor
- `hashmap_reserve_after_fallible_init` -- `reserve()` works after fallible construction

All 26 tests pass (verified on Rust 1.95.0 stable).

## Concerns for the future

1. **Bare `::new()` calls in new code.** Nothing prevents a future contributor from writing `ConcurrentHashMap::new()` in a new module and reintroducing the lazy-init problem. A PR check (e.g., a clippy lint or a simple grep in CI) could prevent regressions: `grep -rn "ConcurrentHashMap::new()" src/`.

2. **`catch_unwind` cannot catch OOM from the global allocator.** `catch_unwind` only catches Rust panics (unwinding). An OOM from `libc::malloc` or `std::alloc::Global` typically aborts the process directly without unwinding. The helpers here protect against panics triggered by papaya's internal assertions, not against true OOM from the OS. The project already uses `cap` for global allocation capping, which is the right layer for that.

3. **`seize::Collector` can still panic at first use.** Papaya's memory reclamation backend (`seize`) lazily initializes its per-thread collector state. Our `with_capacity(1)` trick forces the hash table allocation but does not necessarily exercise the full seize initialization path. If seize panics on a later thread's first `pin()` call, that panic is still caught by the panic hook (since it happens after startup), but it may be surprising. Issue #324 (panic recovery levels) tracks this class of problem.

4. **`with_capacity(1)` is a heuristic.** The capacity passed is a lower bound. Papaya rounds up to the next power of two, so `1` becomes `2`. If papaya changes its internal allocation strategy in a future version, `with_capacity(1)` might no longer force the table allocation. A regression test (`hashmap_put_get` etc.) would catch this because the map would still work correctly, but the safety property (synchronous init) could be silently lost. Periodically reviewing papaya upstream changes is advisable.

5. **The owners of the code should treat this as a pattern, not a panacea.** Wrapping every collection in `catch_unwind` has a non-zero runtime cost (setting up unwind state). For hot-path allocations, consider whether the safety is worth the overhead. The module-level docs in `utils::fallible` include a usage table to guide this decision.


jimezesinachi:
This is great work @gregorian-09, thank you! Can you have a look at the failing tests? I think you need to run `make grpc-update-clients` & `make format`

gregorian-09:
Good afternoon Jim,

Thanks for pointing me to `make grpc-update-clients` and `make format`. I checked the failed jobs, reproduced the failures locally, and pushed the fixes in `5ac4a6c9`.

The Rust job was failing before it reached the tests. It stopped at **Check Rust Client is up to date** because `make grpc-update-rust` changed files under `ahnlich/types/src`.

While checking that, I noticed that the Rust generator walked the generated files without sorting them and stored module names in a `HashSet`. Since neither order is guaranteed, running the generator could write module declarations in a different order and leave CI with a dirty worktree.

I changed the walk to sort by filename and replaced the `HashSet` with a `BTreeSet`, then committed the output from `make grpc-update-rust`. I ran the command multiple times afterward and confirmed that it now produces the same output each time.

I used `make grpc-update-rust` instead of the full `make grpc-update-clients` because the failing check only validates the Rust client. The Go generated-client check is currently disabled, and the Node and Python jobs passed.

The Go job passed its format and lint steps, then failed while running the integration tests. I found two issues in the test setup:

- The helper treated Cargo's `Starting` log message as server readiness. That message can appear before `ahnlich-ai` is listening, so the tests could try to connect too early.
- Every AI test started the server with all default models, even though these tests only need one model. On a fresh runner this caused several large downloads and made the tests fail when the model host rate-limited or interrupted a download.

I changed the helper to wait until the selected TCP port is accepting connections. It also reports process errors and panics instead of continuing to wait.

I added a `SupportedModelsFlag` test option and changed the shared AI test launcher, including the authentication tests, to use only `all-minilm-l6-v2`. I also added the existing `~/.ahnlich` model cache to the Go job so it does not download the same model again on every run.

I also checked the Performance Regression failure. The benchmarks and comparison passed. The only failed step was the step that posts the execution number to the PR:

```text
Resource not accessible by integration
```

Since the benchmark itself passed, I did not change the benchmark workflow in this commit.

I initially tried a broader Rust test-helper change that made each Rust AI test load only its required model. It helped with local nextest runs, but the Rust CI job never reached those tests, so I removed that change before committing.

The final commit contains the deterministic Rust generation fix, the required regenerated/formatted Rust files, and the Go test-startup/model-download fixes.

Verification completed:

- `make grpc-update-rust` repeatedly: stable output
- `make clippy`: passed
- `cargo nextest run -p ahnlich_types -p utils --no-capture`: 37/37 passed
- Full Go AI test package: passed
- Full Go DB test package: passed
- `go vet ./...`: passed
- Go formatting check: passed
- `git diff --check`: passed

I also ran the model-heavy Rust client tests while investigating. Several targeted model cases passed, and the package run reached 16 passing tests. I stopped the remaining first-run model downloads because of the amount of data they required, so I am not claiming a complete full-workspace nextest pass.

Commit:

```text
5ac4a6c9 ci: stabilize generated clients and Go AI tests
```

deven96 (approved):

PR #340, opened 2026-06-03
add Kubernetes (Helm) deployment guide + ahnlich-ai startupProbe for first-run model download 

Iamdavidonuh:


web/ahnlich-web/docs/ahnlich-in-production/kubernetes.md
```diff
@@ -0,0 +1,456 @@
...
+- `kubectl`, configured for the target cluster:
+
+  ```bash
+  kubectl config current-context
+  kubectl get nodes
+  ```
+
+## Get the Charts
+
+There is no chart repository yet, so fetch the charts from GitHub. You only need the
```
deven96, line 40: What would it take us long term to deploy the helm charts to https://artifacthub.io/ ? Probably a next drive to set that up on our CI

deven96 (approved):

PR #341, opened 2026-06-15
Schema Support - Logical Separation for Stores

gregorian-09:
## Schema Support — Logical Separation for Stores

Closes #319

---

Deven raised the idea for schema support in issue #319, and rightfully so. The flat namespace has been a pain point — you cannot have two stores with the same name, which makes multi-tenant setups, environment separation, and even basic organization unnecessarily difficult. You end up with names like `embeddings_dev`, `embeddings_staging`, `embeddings_prod` and before long you are doing mental gymnastics to keep track.

This PR introduces schema-level namespacing. A schema is a lightweight string label that groups stores logically — think of it as a folder.

Before:

```
CREATESTORE dev_embeddings DIMENSION 512
CREATESTORE prod_embeddings DIMENSION 512
```

After:

```
CREATESTORE embeddings DIMENSION 512 SCHEMA dev
CREATESTORE embeddings DIMENSION 512 SCHEMA prod
```

Same name. Different schemas. No collisions.

---

### What changed under the hood

The core data structure went from a flat map to a nested one. Instead of `HashMap<StoreName, Store>`, it is now `HashMap<Schema, HashMap<StoreName, Store>>` in both the DB and AI proxy. The schema is just a string — no configuration, no upfront setup.

New capabilities:

- **DropSchema**: Removes an entire schema and all stores within it. Dropping `"public"` is blocked — that is the default schema where all existing stores reside, and dropping it would break existing setups.
- **ListStores with a filter**: Pass a schema to see only that schema's stores, or omit it to see everything across all schemas.
- **GetStore / DropStore with optional schema**: When no schema is specified, the server searches all schemas. An exact match on `"public"` takes priority for backward compatibility.

Fixed a bug in `PurgeStores` while I was at it — it was counting schemas instead of stores, so the `deleted_count` was wrong when multiple stores existed in the same schema.

### Protobuf

All relevant messages (`CreateStore`, `GetStore`, `DropStore`, `ListStores`) now include an optional `string schema` field. A new `DropSchema` message and matching RPC have been added to both the DB and AI services. SDK stubs for Go, Python, and Node.js have been regenerated.

### CLI

The DSL now supports `SCHEMA` as a keyword in `CREATESTORE` and `DROPSCHEMA` as a top-level command. Existing scripts continue to work unchanged — schema is entirely optional.

### Backward compatibility

Every store created without a schema lives in `"public"`. When no schema is provided, `"public"` is assumed. Existing clients can read, write, and query without any modifications. The only difference is that `ListStores` responses now include a `schema` field if you inspect the payload.

---

### Testing

Integration tests were run across all four codebases:

- Rust DB schema tests: 5 passed
- Rust AI proxy schema tests: 3 passed (limited to one model to keep runtime practical)
- Rust DSL parser: 28 passed
- Go SDK (DB + AI + auth): 41 passed
- Node.js SDK (DB + AI): 15 passed
- Python SDK (DB + AI + auth): 59 passed

The full Rust AI suite (35 tests) requires about 40 minutes and 3.6 GB of model cache. It passes on CI hardware — local runs were constrained by WSL resource limits. The 13 tests that completed before the environment gave out all passed, including the purge fix.

---

### What remains

Two items were deliberately scoped out of this PR:

1. **SDK client wrappers** — The regenerated protobuf stubs include the schema fields, but the hand-written Go, Python, and Node.js client libraries do not yet expose `schema` as a parameter. Raw gRPC callers can use it today. Updating the wrappers is mechanical and better done as a separate change.

2. **Per-schema access control** — As Deven noted in the issue, this is a natural extension. The architecture already supports it since the schema is the outer key — permission checks at that level will be straightforward when the time comes.

There is also a minor issue with the persistence test where a stale `.dat` file from an interrupted run can cause a false negative. Worth addressing in a follow-up.


sdk/ahnlich-client-py/build_fixed.py
deven96, line 1: Not sure where build_fixed is called

deven96:
Thanks for picking this up @gregorian-09!

We rely on the internal representation of a Store to be persistent as we impl `Serialize/Deserialize`. With the introduction of schema keys pointing to internal stores, current ahnlich users with persistence turned on would break upon upgrade.

gregorian-09:
Good catch. The persistence format changed from a flat map to a nested one, so existing persistence files won't deserialize after upgrade.

The fix is a migration in `load_snapshot`: try deserializing the new format first, and if that fails, fall back to the old flat format and wrap it under `"public"`. Something like:

```
fn migrate_snapshot(data: &str) -> Stores {
    // Try new format first
    if let Ok(stores) = serde_json::from_str::<HashMap<Schema, HashMap<StoreName, Store>>>(data) {
        return stores;
    }
    // Fall back to old flat format, wrap in "public"
    let flat: HashMap<StoreName, Store> = serde_json::from_str(data)?;
    let mut stores = HashMap::new();
    stores.insert(Schema::default(), Arc::new(flat.into_iter().collect()));
    stores
}
```

I'll push a commit to handle this. Want to also make sure we bump the persistence file version so future migrations are easier to detect.

deven96:
Any hand written additions (if they even exist) to the client can be done in a different task along with public facing documentation changes to https://ahnlich.dev

deven96:
> Good catch. The persistence format changed from a flat map to a nested one, so existing persistence files won't deserialize after upgrade.
> 
> The fix is a migration in `load_snapshot`: try deserializing the new format first, and if that fails, fall back to the old flat format and wrap it under `"public"`. Something like:
> 
> ```
> fn migrate_snapshot(data: &str) -> Stores {
>     // Try new format first
>     if let Ok(stores) = serde_json::from_str::<HashMap<Schema, HashMap<StoreName, Store>>>(data) {
>         return stores;
>     }
>     // Fall back to old flat format, wrap in "public"
>     let flat: HashMap<StoreName, Store> = serde_json::from_str(data)?;
>     let mut stores = HashMap::new();
>     stores.insert(Schema::default(), Arc::new(flat.into_iter().collect()));
>     stores
> }
> ```
> 
> I'll push a commit to handle this. Want to also make sure we bump the persistence file version so future migrations are easier to detect.

Probably worth introducing two json objects to pin down snapshot persistence loading for AI and DB and catch such issues. Currently what we have is more of a roundtrip test so it wouldn't have caught this

deven96:
Format and lint stage still failing on the PR but I see the file diff has dropped from 100 changed to 64 which is definitely better for review

gregorian-09:
> Probably worth introducing two json objects to pin down snapshot persistence loading for AI and DB and catch such issues. Currently what we have is more of a roundtrip test so it wouldnt have caught this ?

Done. I have committed two fixture files that capture the **actual old flat format** as it existed pre-schema — one for DB (`db/src/tests/fixtures/db_old_flat_snapshot.json`) and one for AI (`ai/src/tests/fixtures/ai_old_flat_snapshot.json`).

Each has a dedicated **read-only** test that loads the fixture file directly, no generation step, no roundtrip:
- `test_migrate_from_committed_db_fixture` in `db/src/tests/server_tests.rs`
- `test_ai_migrate_from_committed_fixture` in `ai/src/tests/migration_test.rs` 

If the serialization format ever shifts, these tests break immediately rather than silently adapting. 6 migration tests total (3 DB + 3 AI), all green. Ready for another review.

gregorian-09:
> Not sure where build_fixed is called, what is it about?

`build_fixed.py` is a standalone development utility for regenerating the Python gRPC client stubs using `betterproto`. It is not called from CI or any automation — it is an alternative to the Makefile `grpc-update-python` target for environments where running the full Makefile workflow is inconvenient. If you prefer the Makefile be the sole generation path, I can drop it from the PR.

deven96:
> > Not sure where build_fixed is called, what is it about?
> 
> `build_fixed.py` is a standalone development utility for regenerating the Python gRPC client stubs using `betterproto`. It is not called from CI or any automation — it is an alternative to the Makefile `grpc-update-python` target for environments where running the full Makefile workflow is inconvenient. If you prefer the Makefile be the sole generation path, I can drop it from the PR.

Yeah we should have one consistent way to do things even if that means running the script from the Makefile. Last push also had autogenerated code changes so feel free to `make format`

ahnlich/ai/src/engine/operations.rs
```diff
@@ -90,6 +92,12 @@ pub async fn create_store(
         predicates.push(AHNLICH_AI_ONE_TO_MANY_INDEX_META_KEY.to_string());
     }
 
+    let schema = params
+        .schema
+        .as_ref()
+        .map(|s| Schema::new(s.clone()))
```
deven96, line 98: Schema will panic on empty string... we don't want to crash but propagate an error accordingly

ahnlich/ai/src/engine/store.rs
```diff
@@ -438,6 +518,28 @@ impl AIStoreHandler {
         Ok(removed)
     }
 
+    /// Drops all stores within a schema and removes the schema.
+    /// Returns the number of stores dropped.
+    #[tracing::instrument(skip(self))]
+    pub(crate) fn drop_schema(&self, schema: &Schema) -> Result<usize, AIProxyError> {
+        if schema.as_str() == Schema::DEFAULT_NAME {
+            return Err(AIProxyError::InvalidArgument(
+                "Cannot drop the default 'public' schema".to_owned(),
```
deven96, line 527: 'public' still hardcoded rather than format with DEFAULT_NAME

ahnlich/types/src/schema.rs
```diff
@@ -0,0 +1,67 @@
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+/// A logical namespace for grouping stores within ahnlich.
+///
+/// Schemas provide multi-tenant isolation and organizational hierarchy
+/// on top of the flat store namespace. Every store belongs to exactly
+/// one schema. The schema `"public"` is the default used when none is
+/// specified, preserving backward compatibility with existing stores.
```
deven96, line 9: We should probably also document case senstivity for schema (given that store names are also case senstive). So "Public" is not the same as "public"

ahnlich/ai/src/engine/store.rs
```diff
@@ -86,10 +90,80 @@ impl AIStoreHandler {
         self.stores = stores_snapshot;
     }
 
+    /// Migrates a flat (pre-schema) AI persistence snapshot into the nested schema format.
+    /// Reads the old `HashMap<StoreName, AIStore>` JSON format and wraps it under `"public"`.
```
deven96, line 94: A better way might be to introduce some `VersionedPersistence` trait concept where we can expressly list out the versions and what we support

```rust
// versioning file under persistence module e.g versioning.rs

#[derive(Error, Debug)]
pub enum VersionError {
    #[error("Persistence file version {file_version} is too new for this ahnlich binary (max supported: {max_version}). Please upgrade ahnlich.")]
    VersionTooNew {
        file_version: u32,
        max_version: u32,
    },
    
    #[error("Persistence file version {file_version} is too old (minimum supported: {min_version}). Migration path removed.")]
    VersionTooOld {
        file_version: u32,
        min_version: u32,
    },
}

/// Trait for versioned persistence types
pub trait VersionedPersistence: Sized + Serialize + for<'de> Deserialize<'de> {
    /// The current version this binary writes
    const CURRENT_VERSION: u32;
    
    /// The minimum version this binary can read
    const MIN_VERSION: u32;
    
    /// Validate version compatibility
    fn validate_version(bytes: &[u8]) -> Result<(), VersionError>;
    
    /// Load from bytes with automatic migration
    fn load_and_migrate(bytes: &[u8]) -> Result<Self, crate::utils::persistence::PersistenceTaskError>;
}
```

DB can implement it then something like

```rust
// In ahnlich/db/src/engine/versioned.rs

pub const DB_CURRENT_VERSION: u32 = 2;
pub const DB_MIN_VERSION: u32 = 1;

// Type aliases
pub type DbStoresV1 = Arc<ConcurrentHashMap<StoreName, Arc<Store>>>;
pub type DbStoresV2 = Arc<ConcurrentHashMap<Schema, Arc<ConcurrentHashMap<StoreName, Arc<Store>>>>>;

/// DB-specific versioned stores
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "db_version")]
pub enum VersionedDbStores {
    #[serde(rename = "1")]
    V1 { 
        stores: DbStoresV1,
    },
    
    #[serde(rename = "2")]
    V2 { 
        stores: DbStoresV2,
    },
    
    // Example future DB versions:
    // #[serde(rename = "3")]
    // V3 { 
    //     stores: Arc<ConcurrentHashMap<Schema, Arc<ConcurrentHashMap<StoreName, Arc<Store>>>>>,
    //     new_db_specific_field: String,
    // },
}

impl VersionedPersistence for VersionedDbStores {
    const CURRENT_VERSION: u32 = DB_CURRENT_VERSION;
    const MIN_VERSION: u32 = DB_MIN_VERSION;
    
    fn validate_version(bytes: &[u8]) -> Result<(), VersionError> {
        #[derive(Deserialize)]
        struct VersionOnly { 
            db_version: Option<String> 
        }
        
        let version = match serde_json::from_slice::<VersionOnly>(bytes) {
            Ok(v) => v.db_version
                .and_then(|s| s.parse::<u32>().ok())
                .unwrap_or(1), // No version tag = V1
            Err(_) => 1,
        };
        
        if version > Self::CURRENT_VERSION {
            return Err(VersionError::VersionTooNew {
                file_version: version,
                max_version: Self::CURRENT_VERSION,
            });
        }
        
        if version < Self::MIN_VERSION {
            return Err(VersionError::VersionTooOld {
                file_version: version,
                min_version: Self::MIN_VERSION,
            });
        }
        
        Ok(())
    }
    
    fn load_and_migrate(bytes: &[u8]) -> Result<Self, PersistenceTaskError> {
        // Validate first
        Self::validate_version(bytes)?;
        
        // Try versioned format
        match serde_json::from_slice::<VersionedDbStores>(bytes) {
            Ok(versioned) => Ok(versioned),
            Err(_) => {
                // Fallback: bare V1 format
                log::warn!("No db_version tag, attempting bare V1 load");
                let v1_stores: DbStoresV1 = serde_json::from_slice(bytes)
                    .map_err(PersistenceTaskError::SerdeError)?;
                Ok(VersionedDbStores::V1 { stores: v1_stores })
            }
        }
    }
}
impl VersionedDbStores {
    /// Get latest version stores (V2)
    pub fn into_latest(self) -> Result<DbStoresV2, String> {
        match self {
            VersionedDbStores::V1 { stores } => {
                log::info!("Migrating DB stores V1 → V2");
                Self::migrate_v1_to_v2(stores)
            }
            VersionedDbStores::V2 { stores } => Ok(stores),
        }
    }
    
    /// Create versioned wrapper for current format
    pub fn current(stores: DbStoresV2) -> Self {
        VersionedDbStores::V2 { stores }
    }
    
    fn migrate_v1_to_v2(v1_stores: DbStoresV1) -> Result<DbStoresV2, String> {
        use utils::fallible;
        
        let v2_stores = fallible::try_new_arc_hashmap()
            .map_err(|e| format!("Migration failed: {}", e))?;
        
        let guard = v2_stores.pin();
        guard.insert(Schema::default(), v1_stores);
        drop(guard);
        
        Ok(v2_stores)
    }
}
```

And then the implementation for ahnlich persistence utils can just simply be 

```rust

impl AhnlichPersistenceUtils for StoreHandler {
    type PersistenceObject = VersionedDbStores;
    fn get_snapshot(&self) -> Self::PersistenceObject {
        VersionedDbStores::current(self.stores.clone())
    }
}
impl StoreHandler {
   
   // We should probably handle MigrationError as a breaking change when starting 
   // or we opt to respect --fail-on-startup-if-persist-load-fails (but that would mean we 
   // can corrupt an old snap shot
    pub fn load_snapshot(bytes: &[u8]) -> Result<Stores, PersistenceTaskError> {
        let versioned = VersionedDbStores::load_and_migrate(bytes)?;
        versioned.into_latest()
            .map_err(|e| PersistenceTaskError::MigrationError(e))
    }
}
```

AIStores would be able to copy the bulk of the implementation and this let's them evolve separately while expressing their schema types and their least supported versions

ahnlich/ai/src/engine/operations.rs
gregorian-09: Yes, thank you for calling my attention to this, Sir. I switched the user-facing call sites from `Schema::new(s.clone())` to `Schema::try_new(s.clone())` which returns a `Result` instead of panicking. The errors bubble up as `InvalidArgument` through gRPC status or server errors, depending on the crate. The one exception is `list_stores`, since that returns a plain `StoreList` (not a `Result`), I just filter out empty strings there rather than silently crashing.

I also added `Schema::try_new` to the types crate so there's a documented non-panicking constructor going forward.

ahnlich/ai/src/engine/store.rs
gregorian-09: Fair point. I have swapped the hardcoded `'public'` for `Schema::DEFAULT_NAME` in both the DB and AI `drop_schema` error messages so they stay in sync if the default ever changes. Thank you for the review.

PR #343, opened 2026-06-16
Schema Support - Logical Separation for Stores

gregorian-09:
Closes #319
Supersedes #341 (source branch feat/schema-support was deleted).

Rebased on latest upstream main with additional fixes:
- Schema::try_new to propagate errors instead of panicking on empty string
- Versioned persistence trait with auto-migration (V1 flat -> V2 nested)
- Hardcoded public references replaced with Schema::DEFAULT_NAME
- Clippy clean, format clean


ahnlich/ai/src/tests/fixtures/ai_old_flat_snapshot.json
deven96, line 1: Probably add one DB and AI json for the new version formats respectively (if that doesn't exist already)

gregorian-09: Done Sir. Added `ai_v2_snapshot.json` and `db_v2_snapshot.json` fixtures under `ahnlich/ai/src/tests/fixtures/` and `ahnlich/db/src/tests/fixtures/`, plus migration tests for both DB and AI loading these V2 snapshots.

gregorian-09:
`tests::server_tests::test_mmap_persistence_performance` is still failing. I’ll do a proper investigation tomorrow and see what’s causing it.

ahnlich/ai/src/engine/operations.rs
```diff
@@ -381,7 +399,7 @@ pub async fn purge_stores(
     parent_id: Option<String>,
 ) -> Result<Del, Status> {
     let store_names: Vec<StoreName> = store_handler
-        .list_stores()
+        .list_stores(None)
```
deven96: So purge stores would only work on public schema?

ahnlich/ai/src/engine/versioned.rs
```diff
@@ -0,0 +1,106 @@
...
+use utils::fallible;
+use utils::persistence::{PersistenceTaskError, VersionedPersistence};
+
+use super::store::{AIStore, AIStores};
+
+pub const AI_CURRENT_VERSION: u32 = 2;
+pub const AI_MIN_VERSION: u32 = 1;
+
+type AiStoresV1 = HashMap<StoreName, AIStore>;
+type AiStoresV1Nested = HashMap<Schema, AiStoresV1>;
```
deven96: Let's expressly call it `AiStoresV2` rather than `AiStoresV1Nested`

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
deven96, line 1: Nice work on these.... maybe generate one using an actual test so that the fields are not empty e.g `id_to_value`, `inner`, `allowed_predicates`, `algorithm_to_index` are currently empty in this test so if anything changes internally in those, this test would still pass

I realise that may be extra work to do but it would be great utility as a regression test for schema changes in general going forward

ahnlich/db/src/engine/versioned.rs
```diff
@@ -0,0 +1,106 @@
...
+use utils::fallible;
+use utils::persistence::{PersistenceTaskError, VersionedPersistence};
+
+use super::store::{Store, Stores};
+
+pub const DB_CURRENT_VERSION: u32 = 2;
+pub const DB_MIN_VERSION: u32 = 1;
+
+type DbStoresV1 = HashMap<StoreName, Store>;
+type DbStoresV1Nested = HashMap<Schema, DbStoresV1>;
```
deven96: Same as the AI comment... we can expressly call it V2

ahnlich/db/src/tests/server_tests.rs
```diff
@@ -4509,15 +4566,470 @@ async fn test_mmap_persistence_performance() {
...
-        mmap_duration,
-        no_mmap_duration
-    );
-
     // Clean up
     let _ = std::fs::remove_file(&mmap_file);
     let _ = std::fs::remove_file(&no_mmap_file);
 }
+
+/// Test: Create stores in different schemas and list them with schema filtering
```
deven96, line 4574: Good additions with the server tests

ahnlich/db/src/tests/server_tests.rs
```diff
@@ -4509,15 +4566,470 @@ async fn test_mmap_persistence_performance() {
...
+        }))
+        .await
+        .expect("GetStore with schema failed")
+        .into_inner();
+
+    assert_eq!(store_info.name, "SchemaGetStore");
+    assert_eq!(store_info.dimension, 4);
+    assert_eq!(store_info.predicate_indices, vec!["tag".to_string()]);
+
+    // GetStore without schema should also find it via fallback
```
deven96: This shouldn't be the case though .... without schema it should default to the "public" schema and hence it wouldn't find this store

deven96: That should be the behaviour of anything we are currently adding that takes optional schema, if schema is not specified, we should default to "public" .... if we allow finding stores across schemas while schema is None, what happens if the same store name exists on multiple schemas?

ahnlich/ai/src/engine/operations.rs
gregorian-09: Yes, that's correct, `purge_stores` now only works on the public schema `(ahnlich/ai/src/engine/operations.rs:401)`. Previously `list_stores(None)` returned stores from all schemas while drop_store with `schema: None` defaulted to public, so non-public schema stores were listed but never actually dropped - this was a bug. I already fixed that by scoping to `list_stores(Some(&Schema::default()))` and passing schema explicitly to `DbDropStore`. If you'd like it to be schema-aware and purge all schemas, that would require adding schema to AiStoreInfo and propagating it through list_stores. Happy to go that route if preferred, let me know.

ahnlich/ai/src/engine/versioned.rs
gregorian-09: Done Sir. Renamed `AiStoresV1Nested` → `AiStoresV2` and `DbStoresV1Nested` → `DbStoresV2` in `ahnlich/ai/src/engine/versioned.rs:16` and `ahnlich/db/src/engine/versioned.rs:16.` They represent the V2 stores structure (nested by schema), just without the serde tag wrapper.

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
gregorian-09: That's a good call. I have gone ahead and done exactly that. The V2 fixture at `db/src/tests/fixtures/db_v2_snapshot.json` now has actual data: two entries with embeddings + metadata, predicates with values wired up, and a built `KdTree` index. One thing I ran into: `StoreKeyId` was using `#[serde(transparent)]` over `u64`, but JSON map keys are always strings and `u64::deserialize` chokes on strings  so the roundtrip was broken. Fixed that by switching to string-based serde for `StoreKeyId` (in `db/src/engine/store.rs`).

The fixture is generated from an `#[ignore]d` test so you can regenerate whenever internal types change, just run cargo test -p db --lib generate_db_v2_fixture -- --ignored. Old V1 snapshots still load fine, and all 67 db lib tests pass

ahnlich/db/src/tests/server_tests.rs
gregorian-09: Yeah, that's a bug. If schema isn't given it should just default to public, searching every schema when names could collide is asking for trouble. I've fixed it though: the store lookup now only checks the default schema, and get_store explicitly takes a schema so there's no ambiguity. Also updated the test to expect the right behaviour.

ahnlich/ai/src/engine/operations.rs
deven96: Yeah I think for consistency sake it would be better to make purge stores schema aware as well

ahnlich/ai/src/tests/aiproxy_test.rs
```diff
@@ -2218,3 +2258,201 @@ async fn test_ai_proxy_embedding_size_mismatch_error() {
...
+    };
+    let response = client
+        .list_stores(tonic::Request::new(message))
+        .await
+        .expect("Failed to list stores")
+        .into_inner();
+    assert_eq!(response.stores.len(), 1);
+    assert_eq!(response.stores[0].name, "AiCustomStore");
+
+    // List stores with no filter - should return all 2 stores
```
deven96: List stores with no filter shouldn't return all stores.... would be best to stick to the behaviour to default to public if schema is None

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
deven96: Good shout with the `id_to_value` StoreKeyId. However, it has worked all this while as u64 map key would be serialized as a string and and also deserialized as such automatically. You can test it with something simple like this

```rust
let test_map: HashMap<u64, String> = HashMap::from_iter([(10, "Random".into())]);
serde_json::to_string_pretty(&test_map).unwrap();
```

It's only an issue when you handroll your json like in this case to account for an older format... so we can get rid of the manual se-deser handling

ahnlich/db/src/tests/server_tests.rs
```diff
@@ -4509,15 +4566,472 @@ async fn test_mmap_persistence_performance() {
...
+        }))
+        .await
+        .expect("ListStores with custom schema failed")
+        .into_inner();
+    assert_eq!(response.stores.len(), 2);
+    let names: Vec<&str> = response.stores.iter().map(|s| s.name.as_str()).collect();
+    assert!(names.contains(&"CustomStore"));
+    assert!(names.contains(&"CustomStore2"));
+
+    // List stores with no schema filter - should return all 3 stores
```
deven96: Probably best to make consistent with all others. Listing all stores across all schemas can become very unwieldy

deven96:
There's been some merged changes from @jimezesinachi with respect to clustering mode for ahnlich so perhaps @jimezesinachi can help work through what would be needed here to make the modifications for logical separation fit with his changes

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
gregorian-09: You were right. The fixture was the problem, not serde. I regenerated `db_v2_snapshot.json` from an actual test so the keys come out as numbers matching `#[serde(transparent)]`, dropped the custom serde, and also refactored the tagged enum loading in `versioned.rs` to peel the version tag first with `std::collections::HashMap` before converting to papaya, that was what was actually breaking the persistence roundtrip, not the transparent derive itself. Clean on both ends now.

ahnlich/ai/src/tests/aiproxy_test.rs
gregorian-09: `list_stores(None)` now defaults to `public` on both DB and AI sides. No more sweeping across all schemas when no schema is given. Same consistent behavior as everything else: if schema is unspecified, it's `"public"`.

gregorian-09:
> There's been some merged changes from @jimezesinachi with respect to clustering mode for ahnlich so perhaps @jimezesinachi can help work through what would be needed here to make the modifications for logical separation fit with his changes

@deven96 @jimezesinachi

So, I went through this diff by diff against everything Jim's clustering touches. writing it up because I want someone to actually check the one thing I'm not 100% sure about, not just rubber-stamp it.

**store_runtime.rs / cluster_mutations.rs** - clean, no diff. enum, `submit_db_command!`, all of Jim's stuff is exactly as he left it.

**handler.rs** - three additions, all additive:

- `drop_schema` handler, copy-pasted the `drop_store` pattern:
```rust
StoreRuntime::Cluster(cluster) => {
    submit_db_command!(Some(cluster), query::DropSchema, params, DbCommand::DropSchema)
}
StoreRuntime::Standalone(store_handler) => {
    operations::drop_schema(store_handler, params)?
}
```
- `list_stores` actually forwards the request body now instead of ignoring it (`list_stores_response(&self.runtime, request.into_inner())`). cluster dispatch inside is untouched, this just lets `schema: None` flow through to standalone.
- `get_store` resolves schema before dispatch, same split as before.
- `new_with_config` swapped `Persistence::load_snapshot` for `load_snapshot_with_migration` to handle the versioned format. standalone-only, cluster has its own snapshot mechanism so this doesn't touch it.

**cluster_queries.rs** - one new param:
```rust
pub(crate) async fn list_stores_response(
    runtime: &StoreRuntime,
    params: query::ListStores,
) -> Result<server::StoreList, tonic::Status> {
    if let Some(cluster) = runtime.cluster() {
        cluster.list_stores(params).await?;
    }
    read_store_handler(runtime, |store_handler| {
        Ok(operations::list_stores(store_handler, params))
    })
}
```
heads up - the cluster branch's return value isn't used, it always falls through to `read_store_handler` after. pretty sure that's pre-existing and not something I introduced, but flagging it because it reads weird on first pass. param only actually lands in `operations::list_stores` via the standalone path.

**replication/mod.rs** - two changes:

- `DropSchema` command, same shape as `DropStore`, nothing surprising.
- `get_snapshot` now does `into_latest()` to turn `VersionedDbStores` into `Stores` before handing it to the cluster, since snapshots are versioned internally but the cluster snapshot type is still plain `Stores`:
```rust
fn get_snapshot(&self) -> Self::Snapshot {
    self.store_handler.get_snapshot().into_latest()
        .expect("snapshot should be at latest version")
}
```
this is the one I want someone else's eyes on. it's an `.expect()` sitting on the replication snapshot path - if a node ever hands back a snapshot that isn't at the latest version, this panics instead of returning an error. probably can't happen given how migration is wired right now, but "probably can't happen" on a path that takes down a node is exactly the kind of thing I don't want to just wave through myself.

`restore_snapshot` - completely untouched.

**operations.rs** - all schema plumbing, nothing clever:
- `resolve_schema()` helper, `Option<String>` to `Schema`, defaults to `public`
- `create_store`/`drop_store`/`list_stores` take schema now, `None` behaves identically to before
- new `drop_schema`, self-contained
- threw in explicit type annotations on `non_linear_indices` (e.g. `StdHashSet<non_linear_index::Index>`) because the compiler asked nicely. no behavior change.

**test files** - `cluster_tests.rs` and `replication_store_tests.rs`, every diff is just `schema: None` added to existing calls. didn't touch any test logic.

so: nothing in Jim's clustering paths actually changed, and everything new degrades to old behavior when schema is `None`. I'm fairly confident in all of it except the `.expect()` in `get_snapshot` - would like one of you to sanity check whether that assumption actually holds before this gets accepted.

deven96:
> > There's been some merged changes from @jimezesinachi with respect to clustering mode for ahnlich so perhaps @jimezesinachi can help work through what would be needed here to make the modifications for logical separation fit with his changes
> 
> @deven96 @jimezesinachi
> 
> So, I went through this diff by diff against everything Jim's clustering touches. writing it up because I want someone to actually check the one thing I'm not 100% sure about, not just rubber-stamp it.
> 
> **store_runtime.rs / cluster_mutations.rs** - clean, no diff. enum, `submit_db_command!`, all of Jim's stuff is exactly as he left it.
> 
> **handler.rs** - three additions, all additive:
> 
> * `drop_schema` handler, copy-pasted the `drop_store` pattern:
> 
> ```rust
> StoreRuntime::Cluster(cluster) => {
>     submit_db_command!(Some(cluster), query::DropSchema, params, DbCommand::DropSchema)
> }
> StoreRuntime::Standalone(store_handler) => {
>     operations::drop_schema(store_handler, params)?
> }
> ```
> 
> * `list_stores` actually forwards the request body now instead of ignoring it (`list_stores_response(&self.runtime, request.into_inner())`). cluster dispatch inside is untouched, this just lets `schema: None` flow through to standalone.
> * `get_store` resolves schema before dispatch, same split as before.
> * `new_with_config` swapped `Persistence::load_snapshot` for `load_snapshot_with_migration` to handle the versioned format. standalone-only, cluster has its own snapshot mechanism so this doesn't touch it.
> 
> **cluster_queries.rs** - one new param:
> 
> ```rust
> pub(crate) async fn list_stores_response(
>     runtime: &StoreRuntime,
>     params: query::ListStores,
> ) -> Result<server::StoreList, tonic::Status> {
>     if let Some(cluster) = runtime.cluster() {
>         cluster.list_stores(params).await?;
>     }
>     read_store_handler(runtime, |store_handler| {
>         Ok(operations::list_stores(store_handler, params))
>     })
> }
> ```
> 
> heads up - the cluster branch's return value isn't used, it always falls through to `read_store_handler` after. pretty sure that's pre-existing and not something I introduced, but flagging it because it reads weird on first pass. param only actually lands in `operations::list_stores` via the standalone path.
> 
> **replication/mod.rs** - two changes:
> 
> * `DropSchema` command, same shape as `DropStore`, nothing surprising.
> * `get_snapshot` now does `into_latest()` to turn `VersionedDbStores` into `Stores` before handing it to the cluster, since snapshots are versioned internally but the cluster snapshot type is still plain `Stores`:
> 
> ```rust
> fn get_snapshot(&self) -> Self::Snapshot {
>     self.store_handler.get_snapshot().into_latest()
>         .expect("snapshot should be at latest version")
> }
> ```
> 
> this is the one I want someone else's eyes on. it's an `.expect()` sitting on the replication snapshot path - if a node ever hands back a snapshot that isn't at the latest version, this panics instead of returning an error. probably can't happen given how migration is wired right now, but "probably can't happen" on a path that takes down a node is exactly the kind of thing I don't want to just wave through myself.
> 
> `restore_snapshot` - completely untouched.
> 
> **operations.rs** - all schema plumbing, nothing clever:
> 
> * `resolve_schema()` helper, `Option<String>` to `Schema`, defaults to `public`
> * `create_store`/`drop_store`/`list_stores` take schema now, `None` behaves identically to before
> * new `drop_schema`, self-contained
> * threw in explicit type annotations on `non_linear_indices` (e.g. `StdHashSet<non_linear_index::Index>`) because the compiler asked nicely. no behavior change.
> 
> **test files** - `cluster_tests.rs` and `replication_store_tests.rs`, every diff is just `schema: None` added to existing calls. didn't touch any test logic.
> 
> so: nothing in Jim's clustering paths actually changed, and everything new degrades to old behavior when schema is `None`. I'm fairly confident in all of it except the `.expect()` in `get_snapshot` - would like one of you to sanity check whether that assumption actually holds before this gets accepted.

_the cluster branch's return value isn't used, it always falls through to read_store_handler after. pretty sure that's pre-existing and not something I introduced_ ... yes that's definitely a bug and not intentional cc @jimezesinachi 

`get_snapshot` is primarily used by the current persistence task to serialize to disk asynchronously at intervals and also by the new RAFT mechanism in `build_snapshot` .... in both cases we are fine to capture the error while logging and propagating accordingly .... if we cannot convert to latest we shouldn't crash but also should not be serializing old formats as is

jimezesinachi:
> > There's been some merged changes from @jimezesinachi with respect to clustering mode for ahnlich so perhaps @jimezesinachi can help work through what would be needed here to make the modifications for logical separation fit with his changes
> 
> @deven96 @jimezesinachi
> 
> So, I went through this diff by diff against everything Jim's clustering touches. writing it up because I want someone to actually check the one thing I'm not 100% sure about, not just rubber-stamp it.
> 
> **store_runtime.rs / cluster_mutations.rs** - clean, no diff. enum, `submit_db_command!`, all of Jim's stuff is exactly as he left it.
> 
> **handler.rs** - three additions, all additive:
> 
> * `drop_schema` handler, copy-pasted the `drop_store` pattern:
> 
> ```rust
> StoreRuntime::Cluster(cluster) => {
>     submit_db_command!(Some(cluster), query::DropSchema, params, DbCommand::DropSchema)
> }
> StoreRuntime::Standalone(store_handler) => {
>     operations::drop_schema(store_handler, params)?
> }
> ```
> 
> * `list_stores` actually forwards the request body now instead of ignoring it (`list_stores_response(&self.runtime, request.into_inner())`). cluster dispatch inside is untouched, this just lets `schema: None` flow through to standalone.
> * `get_store` resolves schema before dispatch, same split as before.
> * `new_with_config` swapped `Persistence::load_snapshot` for `load_snapshot_with_migration` to handle the versioned format. standalone-only, cluster has its own snapshot mechanism so this doesn't touch it.
> 
> **cluster_queries.rs** - one new param:
> 
> ```rust
> pub(crate) async fn list_stores_response(
>     runtime: &StoreRuntime,
>     params: query::ListStores,
> ) -> Result<server::StoreList, tonic::Status> {
>     if let Some(cluster) = runtime.cluster() {
>         cluster.list_stores(params).await?;
>     }
>     read_store_handler(runtime, |store_handler| {
>         Ok(operations::list_stores(store_handler, params))
>     })
> }
> ```
> 
> heads up - the cluster branch's return value isn't used, it always falls through to `read_store_handler` after. pretty sure that's pre-existing and not something I introduced, but flagging it because it reads weird on first pass. param only actually lands in `operations::list_stores` via the standalone path.
> 
> **replication/mod.rs** - two changes:
> 
> * `DropSchema` command, same shape as `DropStore`, nothing surprising.
> * `get_snapshot` now does `into_latest()` to turn `VersionedDbStores` into `Stores` before handing it to the cluster, since snapshots are versioned internally but the cluster snapshot type is still plain `Stores`:
> 
> ```rust
> fn get_snapshot(&self) -> Self::Snapshot {
>     self.store_handler.get_snapshot().into_latest()
>         .expect("snapshot should be at latest version")
> }
> ```
> 
> this is the one I want someone else's eyes on. it's an `.expect()` sitting on the replication snapshot path - if a node ever hands back a snapshot that isn't at the latest version, this panics instead of returning an error. probably can't happen given how migration is wired right now, but "probably can't happen" on a path that takes down a node is exactly the kind of thing I don't want to just wave through myself.
> 
> `restore_snapshot` - completely untouched.
> 
> **operations.rs** - all schema plumbing, nothing clever:
> 
> * `resolve_schema()` helper, `Option<String>` to `Schema`, defaults to `public`
> * `create_store`/`drop_store`/`list_stores` take schema now, `None` behaves identically to before
> * new `drop_schema`, self-contained
> * threw in explicit type annotations on `non_linear_indices` (e.g. `StdHashSet<non_linear_index::Index>`) because the compiler asked nicely. no behavior change.
> 
> **test files** - `cluster_tests.rs` and `replication_store_tests.rs`, every diff is just `schema: None` added to existing calls. didn't touch any test logic.
> 
> so: nothing in Jim's clustering paths actually changed, and everything new degrades to old behavior when schema is `None`. I'm fairly confident in all of it except the `.expect()` in `get_snapshot` - would like one of you to sanity check whether that assumption actually holds before this gets accepted.

@gregorian-09, the version of `list_stores_response` in the actual code in the `main` branch does not look like this at all. Here is what it looks like:

```rust
pub(crate) async fn list_stores_response(
    runtime: &StoreRuntime,
) -> Result<server::StoreList, tonic::Status> {
    if let Some(cluster) = runtime.cluster() {
        cluster
            .raft
            .ensure_linearizable()
            .await
            .map_err(map_linearizable_error)?;
    }

    read_store_handler(runtime, |store_handler| {
        Ok(operations::list_stores(store_handler))
    })
}
```

gregorian-09:
> > > There's been some merged changes from @jimezesinachi with respect to clustering mode for ahnlich so perhaps @jimezesinachi can help work through what would be needed here to make the modifications for logical separation fit with his changes
> > 
> > 
> > @deven96 @jimezesinachi
> > So, I went through this diff by diff against everything Jim's clustering touches. writing it up because I want someone to actually check the one thing I'm not 100% sure about, not just rubber-stamp it.
> > **store_runtime.rs / cluster_mutations.rs** - clean, no diff. enum, `submit_db_command!`, all of Jim's stuff is exactly as he left it.
> > **handler.rs** - three additions, all additive:
> > 
> > * `drop_schema` handler, copy-pasted the `drop_store` pattern:
> > 
> > ```rust
> > StoreRuntime::Cluster(cluster) => {
> >     submit_db_command!(Some(cluster), query::DropSchema, params, DbCommand::DropSchema)
> > }
> > StoreRuntime::Standalone(store_handler) => {
> >     operations::drop_schema(store_handler, params)?
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * `list_stores` actually forwards the request body now instead of ignoring it (`list_stores_response(&self.runtime, request.into_inner())`). cluster dispatch inside is untouched, this just lets `schema: None` flow through to standalone.
> > * `get_store` resolves schema before dispatch, same split as before.
> > * `new_with_config` swapped `Persistence::load_snapshot` for `load_snapshot_with_migration` to handle the versioned format. standalone-only, cluster has its own snapshot mechanism so this doesn't touch it.
> > 
> > **cluster_queries.rs** - one new param:
> > ```rust
> > pub(crate) async fn list_stores_response(
> >     runtime: &StoreRuntime,
> >     params: query::ListStores,
> > ) -> Result<server::StoreList, tonic::Status> {
> >     if let Some(cluster) = runtime.cluster() {
> >         cluster.list_stores(params).await?;
> >     }
> >     read_store_handler(runtime, |store_handler| {
> >         Ok(operations::list_stores(store_handler, params))
> >     })
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > heads up - the cluster branch's return value isn't used, it always falls through to `read_store_handler` after. pretty sure that's pre-existing and not something I introduced, but flagging it because it reads weird on first pass. param only actually lands in `operations::list_stores` via the standalone path.
> > **replication/mod.rs** - two changes:
> > 
> > * `DropSchema` command, same shape as `DropStore`, nothing surprising.
> > * `get_snapshot` now does `into_latest()` to turn `VersionedDbStores` into `Stores` before handing it to the cluster, since snapshots are versioned internally but the cluster snapshot type is still plain `Stores`:
> > 
> > ```rust
> > fn get_snapshot(&self) -> Self::Snapshot {
> >     self.store_handler.get_snapshot().into_latest()
> >         .expect("snapshot should be at latest version")
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this is the one I want someone else's eyes on. it's an `.expect()` sitting on the replication snapshot path - if a node ever hands back a snapshot that isn't at the latest version, this panics instead of returning an error. probably can't happen given how migration is wired right now, but "probably can't happen" on a path that takes down a node is exactly the kind of thing I don't want to just wave through myself.
> > `restore_snapshot` - completely untouched.
> > **operations.rs** - all schema plumbing, nothing clever:
> > 
> > * `resolve_schema()` helper, `Option<String>` to `Schema`, defaults to `public`
> > * `create_store`/`drop_store`/`list_stores` take schema now, `None` behaves identically to before
> > * new `drop_schema`, self-contained
> > * threw in explicit type annotations on `non_linear_indices` (e.g. `StdHashSet<non_linear_index::Index>`) because the compiler asked nicely. no behavior change.
> > 
> > **test files** - `cluster_tests.rs` and `replication_store_tests.rs`, every diff is just `schema: None` added to existing calls. didn't touch any test logic.
> > so: nothing in Jim's clustering paths actually changed, and everything new degrades to old behavior when schema is `None`. I'm fairly confident in all of it except the `.expect()` in `get_snapshot` - would like one of you to sanity check whether that assumption actually holds before this gets accepted.
> 
> @gregorian-09, the version of `list_stores_response` in the actual code in the `main` branch does not look like this at all. Here is what it looks like:
> 
> ```
> pub(crate) async fn list_stores_response(
>     runtime: &StoreRuntime,
> ) -> Result<server::StoreList, tonic::Status> {
>     if let Some(cluster) = runtime.cluster() {
>         cluster
>             .raft
>             .ensure_linearizable()
>             .await
>             .map_err(map_linearizable_error)?;
>     }
> 
>     read_store_handler(runtime, |store_handler| {
>         Ok(operations::list_stores(store_handler))
>     })
> }
> ```

@deven96 @jimezesinachi I re-checked the current diff against `main`, and Jim is right about the list_stores_response path. I'm really sorry about that, Jim. The code on main does not call `cluster.list_stores(params).await?` and then ignore the result. It calls `ensure_linearizable()` in cluster mode, then reads from the local store handler. So there isn’t a fall-through bug in that specific path like I thought. I apologize once again boss.

The re-check did surface the behavior we discussed around omitted schemas. DB `ListStores` `{ schema: None }` was still broader than AI, because it could list stores across schemas. I’ve changed that so an omitted schema defaults to public, matching the intended behavior. The DB schema test has been updated to assert that `schema: None` only returns public stores. I have also adjusted the AI list-store enrichment path to keep that behavior consistent. AI already filters its own stores by schema, but when it fetched DB metadata it was using the no-schema list call. With the DB behavior corrected, that would only look in `public`, even for a custom-schema AI request. The DB client now has a schema-aware list helper, and AI passes the requested schema through when enriching store metadata.

While, for the snapshot path, I also changed the RAFT snapshot hook so snapshot conversion errors are propagated instead of panicking. `get_snapshot` now returns a `Result`, `build_snapshot` forwards that error, and the DB implementation maps conversion failures into `StorageError::IO`. That keeps the node from crashing on snapshot conversion failure while still failing the snapshot operation clearly.

ahnlich/dsl/src/tests/ai.rs
deven96, line 1: Seems we have no DSL implementation and therefore no specific tests for AI/DB that cover the following
- Optional parsing of schema keyword and argument
- DROPSCHEMA query

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
deven96: I'd rather the flat old json snapshot for AI/DB had all the fields populated just like v2

PS: We should include both HNSW index into a store there to cover that too

deven96 (changes requested):

ahnlich/dsl/src/tests/ai.rs
gregorian-09: Thanks, Deven. I've addressed this now.

The DSL now supports the schema forms you called out for both AI and DB: optional `SCHEMA <name>` parsing for `CREATESTORE`, `GETSTORE`, `DROPSTORE`, and `LISTSTORES`, plus `DROPSCHEMA <name>` as a top-level command.

I also added parser tests for both AI and DB so this is covered directly, including the AI DSL test file you pointed at. While wiring it through, I fixed the AI `GETSTORE` schema path and made AI `DropSchema` cascade to the backing DB as documented.

I updated `docs/schema.md` as well so it reflects the current behavior, especially that omitted schema defaults to `public`.

One note on the earlier CI failure: the Rust job failed while trying to download `deven96/all-MiniLM-L12-v2` from Hugging Face and got a `429` rate-limit response. That failure wasn't from the schema/DSL changes.

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
deven96: Awesome work so far.... I think this is what is largely pending

deven96:
I think we are almost there @gregorian-09 

One unresolved comment left

ahnlich/db/src/tests/fixtures/db_old_flat_snapshot.json
gregorian-09: Good evening, Deven. I have addressed that now. The old flat AI and DB snapshot fixtures are now generated from real handlers instead of staying mostly empty. For DB, the flat old fixture now includes inserted values, populated predicate indices, and nonlinear index state. I also updated the DB fixture generation so the store includes both KDTree and HNSW, and tightened the migration tests to assert that those populated fields survive loading/migration. While, for AI, I switched the fixture path to the same generated-fixture approach and fixed the test that was previously overwriting the committed old flat fixture during normal test runs.

deven96 (approved):

deven96:
Approved but I'm giving the benchmarks some time to run as I'd want to be sure we didn't introduce some regrettable performance regression @gregorian-09

deven96:
> Approved but I'm giving the benchmarks some time to run as I'd want to be sure we didn't introduce some regrettable performance regression @gregorian-09

The performances look largely the same so I will go ahead and merge this... Just realized the commenter for the benchmarks is broken though cc @Iamdavidonuh

deven96:
As a follow up @gregorian-09 , we would probably need to update the public facing documentation to show that we support logical separation via schemas in all the commands it was introduced plus the new commands

And then we finally shore up a few additive tests to the autogenerated libraries. Afterwards we should then be ready to cut a new release for the binaries and libraries @Iamdavidonuh

PR #345, opened 2026-06-26
Document schema support and add generated client tests

gregorian-09:
This follows up on the schema implementation merged in #343 and closes #319. The main feature is already in, but Deven pointed out that the public-facing docs and generated-client coverage still needed to catch up. This PR adds schema documentation across the public docs, DB/AI references, CLI docs, and client-library docs, including the important default behavior: when `schema` is omitted, requests target `public`, and `LISTSTORES` does not sweep across every schema.

I also added additive schema lifecycle tests for the generated Go, Python, and Node.js clients on both DB and AI. Those cover creating stores in a custom schema, confirming default list/get remains scoped to `public`, explicitly listing/getting by schema, and dropping a schema. While doing that, I also corrected the stale generated/proto comments that still said unset `ListStores.schema` lists all schemas. Local focused checks pass, and the fork `main` CI is now green.


web/ahnlich-web/docs/components/ahnlich-ai/reference.md
```diff
@@ -32,11 +32,12 @@ INFOSERVER
 
 ## 3. List Stores
 #### Description
-Displays all stores currently created in Ahnlich AI.
+Displays AI stores in the selected schema. If no schema is supplied, only `public` stores are returned.
 
 #### Command
 ```
 LIST STORES
```
deven96: I think this should be LISTSTORES but was originally incorrectly LIST STORES

gregorian-09: I have corrected that. Thank you for calling it out.

web/ahnlich-web/docs/components/ahnlich-ai/reference.md
```diff
@@ -100,7 +113,7 @@ Retrieve the top N most similar vectors to a given raw input.
 GETSIMN 3 WITH "renewable energy storage" USING cosinesimilarity IN article_store WHERE (category != "sports")
```
deven96, line 114: schema support example should be added for all commands that support it

web/ahnlich-web/docs/client-libraries/schema-support.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Schema Support
+---
+
+# Schema Support
+
+Generated clients expose schema support directly from the protobuf API.
+
+For DB and AI, `CreateStore`, `GetStore`, `DropStore`, and `ListStores` accept an optional `schema` field. If `schema` is omitted, the server uses `public`. `ListStores` without a schema returns only stores in `public`.
```
deven96: Ahhh I see... we must have missed schema for all the other commands? 
There's no way to then run GetSimN for non-public stores ... and that makes them not as useful

deven96: Ideally every command that once took only store name should now take an optional schema name

gregorian-09:
Thanks Deven, you were right. I had only carried schema through the store lifecycle commands, but that left the rest of the store-name API less useful because commands like `GETSIMN`, `SET`, `GETPRED`, `GETKEY`, and index operations could not target a non-public schema.

I updated the DB store lookup so store-name operations resolve against the explicit schema instead of implicitly using `public`:

```rust
fn get(&self, store_name: &StoreName, schema: &Schema) -> Result<Arc<Store>, ServerError> {
    let inner_stores = self
        .get_schema(schema)
        .ok_or_else(|| ServerError::StoreNotFound(store_name.clone()))?;

    let inner_guard = inner_stores.guard();
    inner_stores
        .get(store_name, &inner_guard)
        .cloned()
        .ok_or_else(|| ServerError::StoreNotFound(store_name.clone()))
}
```

That is then used by the DB operations that work on an existing store, for example `create_pred_index`, `del_key_in_store`, `get_sim_in_store`, `get_pred_in_store`, `get_key_in_store`, `set_in_store`, and the index drop paths in `ahnlich/db/src/engine/store.rs`.

I made the same correction on the AI side. The important part there is that AI previously checked `public` first and then fell back across every schema. That would have made an omitted schema behave differently from the `public` default we agreed on, so I changed AI lookup to be schema-explicit too:

```rust
pub(crate) fn get(
    &self,
    store_name: &StoreName,
    schema: &Schema,
) -> Result<Arc<AIStore>, AIProxyError> {
    let inner_stores = self
        .get_schema(schema)
        .ok_or_else(|| AIProxyError::StoreNotFound(store_name.clone()))?;

    let inner_guard = inner_stores.guard();
    inner_stores
        .get(store_name, &inner_guard)
        .cloned()
        .ok_or_else(|| AIProxyError::StoreNotFound(store_name.clone()))
}
```

For AI proxy requests that delegate to DB, I also resolve the optional schema once and forward it into the backing DB request. For example, index creation now carries the schema through:

```rust
let schema = resolve_schema(&params.schema)?;
let res = require_db_client(db_client)?
    .create_pred_index(
        DbCreatePredIndex {
            store: params.store,
            predicates: params.predicates,
            schema: Some(schema.to_string()),
        },
        parent_id,
    )
    .await?;
```

I also extended the DSL so the schema clause is available on the commands that operate on a store name, not just create/get/drop/list store. That includes DB and AI `SET`, `GETKEY`, `GETPRED`, `GETSIMN`, `DELKEY`, predicate index commands, and non-linear index commands:

```pest
get_key = { ... ~ store_name ~ (whitespace* ~ schema_clause)? }
get_pred = { ... ~ store_name ~ (whitespace* ~ schema_clause)? }
get_sim_n = { ... ~ store_name ~ (whitespace* ~ schema_clause)? ~ whitespace* ~ (^"where" ~ whitespace* ~ predicate_condition)? }
set_in_store = { ... ~ store_name ~ (whitespace* ~ schema_clause)? }
```

I added focused DSL coverage for both DB and AI schema clauses, including `GETSIMN ... IN store SCHEMA schema WHERE ...`, and updated the docs/client references so the public docs now describe the real behavior: omitted schema means `public`, and `LISTSTORES` does not sweep all schemas.

gregorian-09:
Also, @deven96,  One CI concern I noticed while chasing the Rust failure: the failed test was not a schema assertion failure, it was the runner hitting `No space left on device` while initializing the AI proxy. I’ve reduced the client test setup so the test servers only load the specific models each test needs, instead of loading the full default model list.

Longer term, I think we should also consider trimming the Rust workflow disk pressure. The Rust job restores the Docker image cache, but the Rust test path does not seem to use Docker directly. If that cache is not needed there, removing it or adding an explicit disk cleanup before the full workspace tests would make these model-heavy tests less fragile on GitHub-hosted runners.

web/ahnlich-web/docs/components/ahnlich-ai/reference.md
```diff
@@ -168,8 +200,10 @@ DELETEKEY doc1 IN article_store
 | CREATE STORE my_store INDEXMODEL all-minilm-l6-v2 QUERYMODEL all-minilm-l6-v2 | ```client.create_store("my_store", "all-minilm-l6-v2", "all-minilm-l6-v2")?;``` | ```client.create_store("my_store", index_model="all-minilm-l6-v2", query_model="all-minilm-l6-v2")``` | ```client.CreateStore(ctx, "my_store", "all-minilm-l6-v2", "all-minilm-l6-v2")``` |
 | SET doc1 "The future of AI in healthcare" WITH &#123;"category":"news"&#125; IN article_store | ```client.set("article_store", "doc1", "The future of AI in healthcare", hashmap!{"category"=>"news"})?;``` | ```client.set("article_store", "doc1", "The future of AI in healthcare", {"category":"news"})``` | ```client.Set(ctx, "article_store", "doc1", "The future of AI in healthcare", map[string]string{"category":"news"})``` |
 | DROP STORE article_store | ```client.drop_store("article_store")?;``` | ```client.drop_store("article_store")``` | ```client.DropStore(ctx, "article_store")``` |
+| DROPSCHEMA media | construct generated `DropSchema` request | construct generated `DropSchema` request | construct generated `DropSchema` request |
```
deven96: Not sure this is correct

web/ahnlich-web/docs/components/ahnlich-ai/reference.md
```diff
@@ -168,8 +200,10 @@ DELETEKEY doc1 IN article_store
...
 | GET SIM N 3 WITH "renewable energy storage" USING cosinesimilarity IN article_store WHERE (category != "sports") | ```client.get_sim_n("article_store", "renewable energy storage", 3, "cosine", Some("category!='sports'"))?;``` | ```client.get_sim_n("article_store", "renewable energy storage", 3, "cosine", predicate="category!='sports'")``` | ```client.GetSimN(ctx, "article_store", "renewable energy storage", 3, "cosine", "category!='sports'")``` |
 | GET BY PREDICATE (category = "news") IN article_store | ```client.get_by_predicate("article_store", "category='news'")?;``` | ```client.get_by_predicate("article_store", "category='news'")``` | ```client.GetByPredicate(ctx, "article_store", "category='news'")``` |
+| GET KEY doc1 IN article_store | construct generated `GetKey` request | construct generated `GetKey` request | construct generated `GetKey` request |
```
deven96: This either... the table shows concrete client examples

ahnlich/ai/src/tests/aiproxy_test.rs
deven96, line 1: We should also add schema specific tests for all the commands , not just fixing the existing tests to point to public schema

gregorian-09: Yeah, that’s fair. The schema path was wired through, but I didn’t make that obvious enough in the PR. The AI proxy tests still needed a dedicated non-public schema case for the store-name commands, and the reference table had a couple of placeholder generated-client entries where it should have had real examples. I’ve cleaned that up now: added a focused AI proxy test that runs the schema-aware store commands against a custom schema, and replaced the `DROPSCHEMA` / `GET KEY` placeholders with concrete client examples. That should make the behavior much easier to verify now.

ahnlich/db/src/tests/server_tests.rs
```diff
@@ -4440,6 +4518,7 @@ async fn test_mmap_persistence_performance() {
             query: Some(Query::Set(db_query_types::Set {
                 store: "LargeStore".into(),
                 inputs,
+                schema: None,
             })),
         };
```
deven96, line 4524: Might seem a bit redundant since you have ai tests that cover schema commands and they call db underneath but we should have db tests that provide a tighter harness

web/ahnlich-web/docs/client-libraries/schema-support.md
deven96, line 1: I think each of the client libraries has documentation pages with the existing commands that have been modified and probably need to be reflected

deven96: e.g https://ahnlich.dev/docs/client-libraries/python/request-db/list-stores

deven96 (changes requested):

deven96:
> Also, @deven96, One CI concern I noticed while chasing the Rust failure: the failed test was not a schema assertion failure, it was the runner hitting `No space left on device` while initializing the AI proxy. I’ve reduced the client test setup so the test servers only load the specific models each test needs, instead of loading the full default model list.
> 
> Longer term, I think we should also consider trimming the Rust workflow disk pressure. The Rust job restores the Docker image cache, but the Rust test path does not seem to use Docker directly. If that cache is not needed there, removing it or adding an explicit disk cleanup before the full workspace tests would make these model-heavy tests less fragile on GitHub-hosted runners.

Yes you are correct.... we should remove the docker cache from unused CI, done that here https://github.com/deven96/ahnlich/pull/346/changes

web/ahnlich-web/docs/client-libraries/go/request-db/create-store.md
```diff
@@ -4,6 +4,10 @@ title: Create Store
 
 # Create Store
 
+## Schema
```
deven96, line 7: So what i meant by updating was modifying the actual commands. I will point to them below

web/ahnlich-web/docs/client-libraries/go/request-db/create-store.md
deven96, line 14: Behavior modifications to show schema

deven96, line 86: Including schema into the code snippets as well

deven96 (approved):

Issue #344, opened 2026-06-26
Regression test for the image search example


deven96:
Adding an AI test to ensure the examples from rust image search don't stop working as advertised

PR #352, opened 2026-07-02
feat: add UPSERT command and fix image search similarity algorithm

### LINKED ISSUE #351: Potential Upsert command
There could be utility in providing a single upsert command that reuses predicates to retrieve a single (embedding,metadata) pair to modify

Given that the results of a predicate search can be multiple (embedding,metadata) pairs, we have to return an error whenever that happens otherwise we provide appropriate ways to upsert the embedding,metadata

deven96:
UPSERT Command:
- Add UPSERT to DB and AI services (insert if new, update if exists)
- Returns (inserted, updated) counts in ServerResult
- Add InternalError variant for invariant violations
- Add comprehensive tests and integration tests for Python, Go, Node SDKs
- Add documentation for all SDK languages

Image Search Fix:
- Change similarity algorithm from DotProductSimilarity to CosineSimilarity
- CosineSimilarity is correct for normalized CLIP embeddings
- Add regression tests to prevent future algorithm changes
- Add CI step to run image-search example tests

Fixes #351, #344


ahnlich/ai/src/server/handler.rs
```diff
@@ -433,6 +435,19 @@ impl AiService for AIProxyServer {
         Ok(tonic::Response::new(res))
     }
 
+    #[tracing::instrument(skip_all)]
+    async fn upsert(
```
jimezesinachi, line 438: Since we'd want upserts to be cluster aware, I think it makes sense that we write it with the existing cluster aware configuration. I think we can just move and slightly rewrite the `upsert_internal()` into `engine/operations.rs` and call it here, similar to `drop_pred_index()` below

ahnlich/ai/src/server/handler.rs
```diff
@@ -1134,4 +1162,79 @@ impl AIProxyServer {
     pub fn local_addr(&self) -> IoResult<SocketAddr> {
         self.listener.local_addr()
     }
+
+    async fn upsert_internal(&self, params: Upsert) -> Result<server::Set, AIProxyError> {
```
jimezesinachi: We should move this into operations and slightly rewrite to fit

jimezesinachi (changes requested):
Seems mostly okay, but we probably want to straighten the cluster-awareness of the AI upsert code

ahnlich/ai/src/server/handler.rs
deven96: Fair call

jimezesinachi (approved):
LGTM

PR #365, opened 2026-07-13
Switch to jemalloc

deven96:
Also added an improvement to vector comparisons

For every vector comparison, we computed `StoreKeyId::from(second_vector)` - which hashes the entire 1024-dim vector (computing hash of 1024 floats = 1024 writes to hasher). However, now we pre-compute the `StoreKeyId` once when creating the search list, and pass `(StoreKeyId, EmbeddingKey)` tuples


jimezesinachi (approved):
LGTM

Iamdavidonuh (approved):

PR #369, opened 2026-07-20
feat(wasm-db): add WASM support

deven96:
Make db crate compile for wasm32-unknown-unknown target. Sequential fallbacks for rayon parallel iterators on WASM. SIMD via pulp for 2-4x vector operation speedup.

Create wasm-db crate with protobuf bytes API.
Full CRUD + similarity search + schemas + metadata.

Add WASM build to release workflow.


Iamdavidonuh (approved):

ahnlich/db/src/engine/operations.rs
```diff
@@ -241,3 +262,114 @@ pub fn drop_schema(
     let dropped = store_handler.drop_schema(&schema)?;
     Ok(dropped as u64)
 }
+
+pub fn get_key(
```
jimezesinachi, line 248: If we're putting this here, we want to update it to look like the .set() above, then update the handler.rs calls, so we have a single source of truth

ahnlich/db/src/engine/operations.rs
```diff
@@ -241,3 +262,114 @@ pub fn drop_schema(
...
+                key: embedding_key.as_slice().to_vec(),
+            }),
+            value: Some(Arc::unwrap_or_clone(store_value)),
+        })
+        .collect();
+
+    Ok(server::Get { entries })
+}
+
+pub fn get_pred(
```
jimezesinachi, line 279: Same as above

ahnlich/db/src/engine/operations.rs
```diff
@@ -241,3 +262,114 @@ pub fn drop_schema(
...
+                key: embedding_key.as_slice().to_vec(),
+            }),
+            value: Some(Arc::unwrap_or_clone(store_value)),
+        })
+        .collect();
+
+    Ok(server::Get { entries })
+}
+
+pub fn get_sim_n(
```
jimezesinachi, line 308: Same as above

ahnlich/db/src/engine/operations.rs
```diff
@@ -241,3 +262,114 @@ pub fn drop_schema(
...
+            }),
+            value: Some(Arc::unwrap_or_clone(value)),
+            similarity: Some(similarity),
+        })
+        .collect();
+
+    Ok(server::GetSimN { entries })
+}
+
+pub fn get_store(
```
jimezesinachi, line 346: Same as above

ahnlich/db/src/engine/operations.rs
```diff
@@ -173,6 +176,8 @@ pub fn drop_store(
 
 pub fn set(store_handler: &StoreHandler, params: query::Set) -> Result<StoreUpsert, ServerError> {
```
jimezesinachi, line 176: We would want the others to look like this one

jimezesinachi (changes requested):
Mostly, LGTM, just a minor rewrite to prevent future drift between the operations.rs and handler.rs impls

jimezesinachi (approved):
LGTM

PR #370, opened 2026-07-20
feat(db): add leader forwarding of cluster-aware requests

jimezesinachi:


ahnlich/db/src/server/cluster_mutations.rs
```diff
@@ -56,29 +65,43 @@ async fn submit_raw_db_command(
...
+    Fut: std::future::Future<Output = Result<tonic::Response<T>, tonic::Status>>,
 {
     let command_name = command_name!(Q);
     let command = command_builder(params.encode_to_vec());
 
-    match submit_raw_db_command(cluster, command_name, command).await? {
-        DbResponse::Bytes(bytes) => bitcode::deserialize(bytes.as_slice()).map_err(|err| {
+    match submit_raw_db_command(cluster, command_name, command).await {
+        Err(ClientWriteFailure::ForwardToLeader(leader_addr)) => {
+            let cluster = cluster.expect("cluster was required before forwarding");
```
deven96: Shouldn't submit raw_db_command in the `ClientWriteFailure::ForwardToLeader` then return some reference to the cluster rather than expecting?

ahnlich/db/src/server/handler.rs
```diff
@@ -383,15 +458,23 @@ impl DbService for Server {
         &self,
         request: tonic::Request<query::DropSchema>,
     ) -> std::result::Result<tonic::Response<server::Del>, tonic::Status> {
+        let metadata = request.metadata().clone();
```
deven96: I think we can use into_parts for these rather than cloning full metadata but I'm unsure... if it works let's do everywhere

deven96 (approved):

ahnlich/db/src/server/cluster_mutations.rs
jimezesinachi: I see what you're saying, good catch, fixing now

deven96 (approved):

PR #378, opened 2026-08-03
feat(replication): add state machine snapshot durability

jimezesinachi:


ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -90,10 +90,45 @@ pub trait StateMachineHandler<C: RaftTypeConfig>: Send + Sync + 'static {
...
 // snapshot bookkeeping.
 
-#[derive(Debug, Clone)]
-struct StoredSnapshot<C: RaftTypeConfig> {
-    meta: SnapshotMeta<C::NodeId, C::Node>,
-    data: Vec<u8>,
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct PersistedSnapshot<C: RaftTypeConfig> {
+    pub meta: SnapshotMeta<C::NodeId, C::Node>,
+    pub data: Vec<u8>,
```
deven96, line 96: I think we can Arc this as it's potentially very large and would be awful to clone down the line

ahnlich/replication/src/storage/state_machine.rs
```diff
@@ -185,7 +252,7 @@ where
     C::SnapshotData: From<Cursor<Vec<u8>>>,
 {
     async fn build_snapshot(&mut self) -> Result<Snapshot<C>, StorageError<C::NodeId>> {
-        let (snapshot, meta) = {
+        let (snapshot, meta, snapshot_idx) = {
```
deven96, line 255: Then snapshot here can be an `Arc<Vec<u8>>`, cheaper to clone the Arc

ahnlich/replication/src/storage/state_machine.rs
jimezesinachi: Good point to consider. I traced `PersistedSnapshot`’s ownership path, and making data an `Arc<Vec<u8>>` would not remove the copy: openraft currently requires an owned `Cursor<Vec<u8>>` for `SnapshotData`, and the state machine must retain the bytes for its current persisted snapshot. We would still need two independently owned payloads, just copying at a later conversion point. Plus we currently only clone `PersistedSnapshot` in the in-memory snapshot store, not the RocksDB implementation. The RocksDB implementation serializes the borrowed snapshot and, on load, deserializes a fresh owned value, so it does not call `PersistedSnapshot::clone()` either

ahnlich/replication/src/storage/state_machine.rs
jimezesinachi: Same limitation applies here: the snapshot returned to openraft must be an owned `Cursor<Vec<u8>>`, so an `Arc` cannot be propagated through the current snapshot-data contract without copying it back into a `Vec`. A genuinely shared snapshot buffer would mean introducing a custom `SnapshotData` impl, and updating the receive/transport path for DB and AI. That is valid with openraft, but substantially wider than this PR, so I’d rather treat it as a performance follow-up

deven96: Alright then, could you please open an issue to track this. Approving the current PR

deven96 (approved):

PR #383, opened 2026-08-06
Ahnlich MCP

### LINKED ISSUE #373: Ahnlich MCP
An MCP server that exposes Ahnlich as tools any MCP-compatible agent can discover and use (Claude Desktop, Cursor, VS Code Copilot, custom agents). Each tool maps to one Ahnlich operation. Agents compose these into workflows.

Potential Use cases:
1. Agent working memory: agent stores intermediate findings during multi-step tasks, searches them later. No pre-built pipeline needed.
2. Local file organization: agent reads a user's Downloads folder with its own filesystem access, stores descriptions in Ahnlich through the MCP server, then searches semantically ("find that ML paper I downloaded last month").
3. Personal knowledge base: agent indexes notes and research as the user shares them. The store becomes searchable memory over time.

osobotu:
This PR adds an MCP server for Ahnlich. Fixes #373 

```mermaid
flowchart LR
    Agent["AI agent"] -->|"stdio"| MCP["ahnlich-mcp"]
    MCP -->|"DB profile: embeddings"| DB["ahnlich-db"]
    MCP -->|"AI profile: raw text"| AI["ahnlich-ai"]
    AI -->|"embeddings"| DB
```

- Adds `db` and `ai` profiles with profile-specific tool schemas.
- Exposes tools for store management, storage, similarity search, metadata filtering, and predicate indexes.
- Adds environment and CLI configuration plus a `doctor` connectivity command.
- Uses `uv` with Python 3.11 and the monorepo Python SDK.
- Separates tools, server setup, configuration, and backend communication.
- Centralizes shared backend behavior while preserving AI- and DB-specific logic.
- Adds unit, real-service integration, and MCP stdio tests.
- Adds CI with dependency/model caching and Docker-based integration testing.
- Documents installation, profiles, configuration, architecture, and development workflows.


wizzywit:
After going through this implementation, there are some standards or optimization gaps that I think would be worth while to be implemented for this feature.

1. No tool annotations. drop_store and delete_by_metadata are irreversible, and ping/list_stores/server_info/get_by_metadata are read-only, but nothing is declared. FastMCP supports mcp.tool(annotations=ToolAnnotations(readOnlyHint=..., destructiveHint=..., idempotentHint=...)) — clients use these to gate confirmation prompts. This is the single highest-value addition for a server whose tools can wipe a store.

2. Errors are returned as successful results. error_response returns {"status": "error", ...} as a normal tool result, so isError is never set. The dict is LLM-legible, which is the important part, but protocol-level clients can't distinguish failure. Consider raising ToolError (or setting the error flag) while keeping the structured hint payload.

3. No read-only mode. An AHNLICH_MCP_READ_ONLY=1 that suppresses registration of drop_store / delete_by_metadata / drop_predicate_index would be cheap and is a common pattern for database MCP servers.

4. get_by_metadata and list_stores are unbounded. A predicate matching 100k entries returns all of them. Add a limit (with a sane default) and report truncation.

mcp/src/ahnlich_mcp/backends/base.py
```diff
@@ -0,0 +1,661 @@
...
+            response.entries
+        )
+
+        if sort_descending:
+            results.sort(
+                key=lambda result: result["similarity"],
+                reverse=True,
+            )
+
+        return results
```
wizzywit: Wrong result ordering for euclidean.

This sorts by the similarity field descending, but that field isn't always a similarity. In ahnlich/similarity/src/distance.rs:118, LinearAlgorithm::EuclideanDistance => euclidean_distance(a, b).into() yields a Score::Distance, where lower is better. So algorithm="euclidean" on the AI profile (the only caller passing sort_descending=True) returns matches worst-first.

The server already ranks entries, and the AI proxy preserves that order through its rayon unzip (ahnlich/ai/src/server/handler.rs:395-407), so no client-side sort is needed for any algorithm. Note also that applying it to AI but not DB means the same query returns different orderings per profile.

```
    async def _execute_similarity_search(
        self,
        *,
        request: Any,
        store_name: str,
        metadata_filter_applied: bool,
    ) -> list[dict[str, Any]]:
        """Run a get_sim_n request.

        Entries are returned in the order the server ranked them. Do not
        re-sort here: for EuclideanDistance the `similarity` field carries
        a distance (lower is better), while cosine and dot product carry a
        similarity (higher is better), so no single client-side sort key is
        correct for every algorithm.
        """
        response = await self._call(
            "get_sim_n",
            request,
            store_name=store_name,
            predicate_operation=metadata_filter_applied,
        )

        return self._format_search_entries(
            response.entries
        )
```

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+            model_params={},
+        )
+
+        return await self._execute_similarity_search(
+            request=request,
+            store_name=store_name,
+            metadata_filter_applied=(
+                metadata_filter is not None
+            ),
+            sort_descending=True,
```
wizzywit: Remove alongside the `base.py` change above.

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if not query.strip():
+            raise ValueError(
+                "query must not be empty"
+            )
+
+        if metadata_filter is not None:
+            condition = self._build_condition(
+                metadata_filter
+            )
```
wizzywit: Dead work: `_prepare_similarity_search` (line 158) already returned `condition` built from this same `metadata_filter`. Recomputing it here duplicates the validation and invites the two paths to drift.

mcp/src/ahnlich_mcp/config.py
```diff
@@ -0,0 +1,121 @@
...
+
+        model = os.getenv(
+            "AHNLICH_AI_MODEL",
+            "all-minilm-l6-v2",
+        ).strip()
+
+        if not model:
+            raise ConfigurationError(
+                "AHNLICH_AI_MODEL cannot be empty"
+            )
```
wizzywit: Unsupported model names produce a traceback rather than a usage error.

This only checks the string is non-empty; the real validation lives in `AIBackend.__init__` (backends/ai.py:52-62) and raises a plain `ValueError`. Since `main` catches only ConfigurationError (server.py:199-202), `AHNLICH_AI_MODEL=bge-base-en-v1.5 uv run ahnlich-mcp` exits with an unhandled traceback.

Suggest moving the registry into a dependency-free ahnlich_mcp/models.py (importable from both `config` and `backends.ai` without a cycle) and validating here:

```
    @staticmethod
    def _read_model() -> str:
        model = os.getenv(
            "AHNLICH_AI_MODEL",
            DEFAULT_TEXT_MODEL,
        ).strip()

        if not model:
            raise ConfigurationError(
                "AHNLICH_AI_MODEL cannot be empty"
            )

        if model not in TEXT_MODELS:
            supported = ", ".join(sorted(TEXT_MODELS))

            raise ConfigurationError(
                f"Unsupported AI model {model!r}. "
                f"Supported models: {supported}"
            )

        return model

```
Keep the guard in `AIBackend.__init__` as defence-in-depth for direct construction.

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+
+from ahnlich_mcp.backends.base import (
+    AlgorithmName,
+    BaseBackend,
+)
+
+
+AI_MODELS = {
+    "all-minilm-l6-v2": AiModel.ALL_MINI_LM_L6_V2,
+}
```
wizzywit: Only 1 of 12 `AiModel` variants is reachable (sdk/ahnlich-client-py/ahnlich_client_py/grpc/ai/models/__init__.py:11-23). The other text embedders are drop-in here — names below match ahnlich-ai's `--supported-models` values (ahnlich/ai/src/cli/server.rs:28-53) so the same string works in both places:

```
TEXT_MODELS: dict[str, AiModel] = {
    "all-minilm-l6-v2": AiModel.ALL_MINI_LM_L6_V2,
    "all-minilm-l12-v2": AiModel.ALL_MINI_LM_L12_V2,
    "bge-base-en-v1.5": AiModel.BGE_BASE_EN_V15,
    "bge-large-en-v1.5": AiModel.BGE_LARGE_EN_V15,
    "jina-embeddings-v2-base-code": (
        AiModel.JINA_EMBEDDINGS_V2_BASE_CODE
    ),
}
```

The multimodal variants (`resnet-50`, `clip-vit-b32-image`, `buffalo-l`, `sface-yunet`, `clap-audio`) legitimately don't fit a `raw_string-only` tool surface — worth a comment saying so rather than leaving the omission implicit. Please also note in the README that the selected model must appear in ahnlich-ai's --supported-models; the bundled compose starts only `all-minilm-l6-v2`, so picking another fails at `create_store` with a confusing server-side error.

mcp/src/ahnlich_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+
+    mcp = FastMCP(
+        name="ahnlich-mcp",
+        instructions=instructions_for(settings.profile),
+        json_response=True,
+        lifespan=lifespan,
+    )
+
+    for tool in build_tools(backend):
+        mcp.tool()(tool)
```
wizzywit: No tool annotations are declared. drop_store, delete_by_metadata, and drop_predicate_index are irreversible; ping, server_info, list_stores, get_by_metadata, and similarity_search are read-only. MCP clients use these hints to decide whether to prompt the user before invoking, so on a server that can wipe a store this is the highest-value addition.

Keying off tool.__name__ avoids changing build_tools' signature, so existing tests keep passing:

```
    for tool in build_tools(backend):
        mcp.tool(
            annotations=TOOL_ANNOTATIONS.get(tool.__name__),
        )(tool)
```

with a TOOL_ANNOTATIONS: dict[str, ToolAnnotations] in tools.py next to TOOL_ORDER. openWorldHint=False throughout (every tool talks only to the one configured instance); store_entries and create_predicate_index can be idempotentHint=True since the underlying set/create_pred_index are upserts.

A test_every_tool_declares_annotations asserting set(TOOL_ANNOTATIONS) == set(TOOL_ORDER) would stop a future tool being added without a blast-radius decision, and an assertion over session.list_tools() in test_mcp_stdio.py would confirm the hints reach the wire.

mcp/src/ahnlich_mcp/server.py
```diff
@@ -0,0 +1,214 @@
...
+                )
+
+            yield
+        finally:
+            await backend.close()
+
+    mcp = FastMCP(
+        name="ahnlich-mcp",
+        instructions=instructions_for(settings.profile),
+        json_response=True,
```
wizzywit: json_response is a streamable-HTTP option and is inert under transport="stdio" (server.py:210). Remove it, or add HTTP transport support so it means something.

mcp/src/ahnlich_mcp/backends/db.py
```diff
@@ -0,0 +1,221 @@
...
+    service_name = "ahnlich-db"
+    query_module = db_query
+    stub_type = db_service.DbServiceStub
+
+    def _entry_identity(self, entry: Any) -> dict[str, Any]:
+        return {
+            "embedding": list(
+                entry.key.key
+            )
+        }
```
wizzywit: Returning the full vector by default is expensive for the caller: top_k=5 against a 1536-dim store puts ~7700 floats into the model's context, and they're vectors the agent supplied in the first place. Keeping dimension preserves the ability to sanity-check what was stored:

```
    def _entry_identity(
        self,
        entry: Any,
        *,
        include_embeddings: bool = False,
    ) -> dict[str, Any]:
        embedding = entry.key.key

        if not include_embeddings:
            return {"dimension": len(embedding)}

        return {"embedding": list(embedding)}
```

That needs include_embeddings threaded through _format_entry / _format_search_entries / _execute_similarity_search / get_by_metadata in base.py, AIBackend._entry_identity accepting and ignoring the keyword, and the DB similarity_search tool exposing it as include_embeddings: bool = False.

mcp/src/ahnlich_mcp/tools.py
```diff
@@ -0,0 +1,420 @@
...
+    async def get_by_metadata(
+        store_name: str,
+        filter: dict[str, str],
+    ) -> ToolResponse:
+        """Retrieve entries matching indexed metadata."""
+        try:
+            return await backend.get_by_metadata(
+                store_name=store_name,
+                metadata_filter=filter,
+            )
```
wizzywit, line 184: get_by_metadata is unbounded — a predicate matching 100k entries returns all of them straight into the model's context. Same applies to list_stores (line ~110). Suggest a limit: int = 50 with an explicit truncation marker in the response so the agent knows results were cut.

mcp/src/ahnlich_mcp/backends/base.py
```diff
@@ -0,0 +1,661 @@
...
+            raise ValueError(
+                "top_k must be an integer"
+            )
+
+        if top_k < 1:
+            raise ValueError(
+                "top_k must be at least 1"
+            )
+
+        return top_k
```
wizzywit, line 505: No upper bound on top_k, so top_k=10_000_000 is forwarded to the server. A ceiling (1024 is reasonable) belongs next to the top_k < 1 check.

mcp/src/ahnlich_mcp/backends/base.py
```diff
@@ -0,0 +1,661 @@
...
+        )
+
+        response = await self._call(
+            "create_pred_index",
+            self.query_module.CreatePredIndex(
+                store=store_name,
+                predicates=validated_keys,
+            ),
+            store_name=store_name,
+        )
```
wizzywit, line 351: **NON-BLOCKING**: Inconsistent with the three sibling predicate operations (lines 280, 304, 351), which all pass predicate_operation=True. As written, create_pred_index errors fall through to generic AhnlichError and lose the index hint.

```
            store_name=store_name,
            predicate_operation=True,
        )
```

mcp/src/ahnlich_mcp/backends/base.py
```diff
@@ -0,0 +1,661 @@
...
+        if (
+            predicate_operation
+            and "predicate" in lowered_message
+            and (
+                "index" in lowered_message
+                or "not found" in lowered_message
+                or "does not exist" in lowered_message
+            )
+        ):
+            return PredicateIndexNotFoundError(message)
```
wizzywit, line 466: **NON-BLOCKING**: Classifying errors by substring match on the server's message text breaks silently the moment a Rust-side string is reworded, and nothing in the Rust crates is contractually bound to these words. At minimum add a comment pointing at where the messages originate, and pin the real strings in an integration test so a rewording fails CI rather than degrading quietly to a generic error.

mcp/src/ahnlich_mcp/tools.py
```diff
@@ -0,0 +1,420 @@
...
+            "endpoint": f"{backend.host}:{backend.port}",
+            "available": available,
+        }
+
+        if not available:
+            result["suggested_action"] = (
+                "Start the configured Ahnlich service and verify its address."
+            )
+
+        return result
```
wizzywit, line 129: **NON-BLOCKING**: ping is the only tool without a try/except. Backend.ping swallows AhnlichError (base.py:117-125) but not ValueError or an unexpected AttributeError, so those escape as protocol-level errors while every other tool returns a structured payload. Wrapping it in error_response would make the surface uniform.

mcp/src/ahnlich_mcp/tools.py
```diff
@@ -0,0 +1,420 @@
...
+
+        return {
+            "status": "dropped",
+            "store_name": store_name,
+            "deleted": deleted,
+        }
+
+    async def get_by_metadata(
+        store_name: str,
+        filter: dict[str, str],
```
wizzywit: **NON-BLOCKING**: filter shadows the builtin. It becomes the wire-level parameter name so it isn't purely cosmetic, but `metadata_filter` is what the backend layer already calls it and would read better in the generated schema. Applies to all five tools taking this argument.

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+    def _entry_identity(self, entry: Any) -> dict[str, Any]:
+        key = entry.key
+
+        return {
+            "content": (
+                key.raw_string
+                if key is not None
+                else ""
+            )
+        }
```
wizzywit: **NON-BLOCKING**: Falling back to "" when key is None silently turns a malformed response into an entry with empty content, which the agent can't distinguish from a genuinely empty record. Consider surfacing it (skip the entry, or mark it) rather than papering over it.

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _build_set_request(
+        self, *,
+        store_name: str,
+        entries: list[dict[str, Any]],
+    ) -> ai_query.Set:
+        return ai_query.Set(
+            store=store_name,
+            inputs=self._build_entries(entries),
+            preprocess_action=PreprocessAction.NoPreprocessing,
```
wizzywit: NON-BLOCKING: PreprocessAction.NoPreprocessing is hardcoded for both set and get_sim_n, so any input exceeding the model's token limit fails server-side with no way for the caller to opt into truncation. Worth exposing as a tool parameter or at least documenting the constraint in the README.

mcp/src/ahnlich_mcp/backends/ai.py
```diff
@@ -0,0 +1,242 @@
...
+            )
+
+        request = ai_query.GetSimN(
+            store=store_name,
+            search_input=StoreInput(raw_string=query),
+            condition=condition,
+            closest_n=top_k,
+            algorithm=algorithm_value,
+            preprocess_action=(
+                PreprocessAction.NoPreprocessing
```
wizzywit: Same as https://github.com/deven96/ahnlich/pull/383#discussion_r3748269266

wizzywit (commented):
Overall the implementation looks solid. 

- Transport errors are translated into domain exceptions carrying an actionable suggested_action, so a failing tool call tells the agent what to do rather than leaking a gRPC status.
- The server stays up when Ahnlich is unreachable (server.py:52-77) instead of dying at startup — right call for a stdio server whose client owns the process lifecycle.
- doctor is a genuinely useful operator affordance.
- Real end-to-end coverage through ClientSession in tests/integration/test_mcp_stdio.py, not just unit mocks.
- Reuses the in-repo sdk/ahnlich-client-py via [tool.uv.sources] rather than vendoring protobufs.

mcp/src/ahnlich_mcp/backends/base.py
osobotu: Preserved the ranking returned by Ahnlich instead of sorting in the MCP backend.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: Removed the AI profile's sort_descending argument.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: Removed the duplicate metadata condition construction.

mcp/src/ahnlich_mcp/config.py
osobotu: Moved the model registry into ahnlich_mcp/models.py and now validate AHNLICH_AI_MODEL while loading settings. Unsupported models produce a CLI usage error.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: Added the supported raw-text embedding models and documented in the readme that the selected model must also be enabled by Ahnlich-AI. The bundled Compose configuration continues to enable only all-minilm-l6-v2 by default.

mcp/src/ahnlich_mcp/server.py
osobotu: Added a complete TOOL_ANNOTATIONS registry and registered each tool with its annotations. Also added unit test coverage that prevents new tools from being added without annotations, and stdio coverage verifies the hints reach MCP clients.

mcp/src/ahnlich_mcp/server.py
osobotu: Removed json_response because the server currently uses stdio transport. In the future, if we include streamable-HTTP, this will be added.

mcp/src/ahnlich_mcp/backends/db.py
osobotu: DB search and metadata results now omit full vectors by default and return their dimension instead. Callers can explicitly set include_embeddings=true when the vector values are required.

mcp/src/ahnlich_mcp/tools.py
osobotu: list_stores and get_by_metadata now return a bounded response envelope like so:
`{
"results": ..., 
"truncated": ... }`
containing results and a truncation. Invalid or excessive limits are rejected before querying Ahnlich.

mcp/src/ahnlich_mcp/backends/base.py
osobotu: Added a maximum top_k of 1024 with validation before forwarding to the server.

mcp/src/ahnlich_mcp/backends/base.py
osobotu: Classified create_predicate_index as a predicate operation so predicate-index errors retain their actionable hint.

mcp/src/ahnlich_mcp/backends/base.py
osobotu: Added a comment in code stating that predicate-index classification depends on the message defined in ahnlich/db/src/errors.rs. Improved the classifier around the real message markers and added unit and real-service integration coverage that maps to the current server string, if it changes, the tests will fail and can be resolved.

mcp/src/ahnlich_mcp/tools.py
osobotu: Wrapped unexpected ping exceptions using the same tool-error response as the other tools.

mcp/src/ahnlich_mcp/tools.py
osobotu: Renamed the public filter parameter to metadata_filter across similarity search, metadata retrieval, and metadata deletion.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: AI entries without a key now raise an explicit AhnlichResponseError instead of silently returning empty content.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: Added a preprocessing option to the AI store_entries and similarity_search tools. Callers can select none or truncate, which maps to Ahnlich's NoPreprocessing and ModelPreprocessing actions respectively.

mcp/src/ahnlich_mcp/backends/ai.py
osobotu: Resolved above

mcp/src/ahnlich_mcp/config.py
```diff
@@ -0,0 +1,148 @@
...
+        ).strip()
+
+        if not model:
+            raise ConfigurationError("AHNLICH_AI_MODEL cannot be empty")
+
+        if model not in TEXT_MODELS:
+            supported = ", ".join(sorted(TEXT_MODELS))
+            raise ConfigurationError(
+                f"Unsupported AI model {model!r}."
+                f"Supported models: {supported}"
```
wizzywit, line 103: **NON-BLOCKING** (introduced by the fix for https://github.com/deven96/ahnlich/pull/383#discussion_r3748089746 — the validation itself is correct):

Missing separator in the implicit string concatenation, so the two sentences run together:

```
AHNLICH_AI_MODEL=bogus uv run ahnlich-mcp
ahnlich-mcp: error: Unsupported AI model 'bogus'.Supported models: all-minilm-l12-v2, all-minilm-l6-v2, bge-base-en-v1.5, bge-large-en-v1.5, jina-embeddings-v2-base-code
```

```suggestion
                f"Unsupported AI model {model!r}. "
                f"Supported models: {supported}"
```

I scanned the rest of `mcp/src/` for the same pattern — this is the only occurrence, every other multi-line message keeps its trailing space.

While you're in this function: line 105 is whitespace-only. Same for `backends/base.py:380,475,580`. A `ruff format --check` step in `mcp_test.yml` would keep these from accumulating (raised in the review summary, not blocking).

deven96 (approved):

wizzywit (approved):

mcp/docker-compose.yml
```diff
@@ -46,4 +53,44 @@ services:
...
+        condition: service_healthy
+
+  mcp-ai:
+    <<: *mcp
+    profiles: ["ai"]
+    command: ["--profile", "ai"]
+    environment:
+      AHNLICH_AI_HOST: ahnlich_ai
+      AHNLICH_AI_PORT: "1370"
+      AHNLICH_AI_MODEL: "${AHNLICH_AI_MODEL:-all-minilm-l6-v2}"
```
deven96, line 77: Shouldn't this be MODELS?

deven96 (approved):
We

PR #384, opened 2026-08-12
feat(replication): add cluster & node identity framework

jimezesinachi:
## Summary

This PR strengthens DB cluster membership and admission so the existing DB Raft implementation is safe to build AI replication on top of.

DB replication already routes and replicates writes, but it previously assumed that nodes were compatible, reachable through their bind addresses, and safe to promote once they joined. Those assumptions are too weak for AI replication, where a joining node will also need to prove that it can safely replay AI state and serve the same model-backed stores.

## What Changed

- Persist a generated, stable node ID in Raft storage instead of deriving it from `--cluster-addr`.
- Persist a generated cluster identity with replication-protocol and state-machine-format versions.
- Add `--cluster-advertise-addr` for Raft/admin traffic and `--cluster-leader-forwarding-advertise-addr` for follower-to-leader DB request forwarding.
  - Both default to their corresponding bound address.
  - They are only needed when the bound address is not reachable by peers, such as `0.0.0.0` or a container-local address.
- Replace the raw `AddLearner` / client-provided `ChangeMembership` join flow with:
  1. `AdmitLearner`: leader validates the candidate’s node ID, endpoints, and format compatibility, then adds it as a learner and waits for catch-up.
  2. `PromoteLearner`: leader promotes that known learner using its current membership view.
- Reject incompatible replication protocol or state-machine format versions both:
  - locally, before a joining node persists cluster identity;
  - on the leader, before it admits the candidate.

## Why This Is Needed

Raft guarantees consensus only when members interpret the replicated log and snapshots the same way. A node with an incompatible replication protocol or state-machine format must not join, otherwise replay can fail after it becomes part of the cluster.

The address split fixes another operational gap: `0.0.0.0` is valid for a server to bind to, but it is not a meaningful endpoint for another node to dial. Raft transport and DB leader forwarding each need a real advertised address when the local bind address is not reachable.

Finally, membership promotion is now leader-owned. A joining process no longer reads membership, constructs its own voter set, and asks the leader to apply it. It can only ask to be admitted and then promoted after the leader has validated it and OpenRaft has caught it up.

## Why This Unblocks AI Replication

AI replication needs the same foundation, plus AI-specific admission checks:

- supported model and effective model configuration;
- compatibility of AI state-machine format;
- binding to the intended DB cluster.

This PR establishes the place and flow for those checks: a durable cluster contract, a candidate payload, leader-side validation, learner catch-up, and explicit promotion. AI replication can extend that admission validation without inventing a second membership model.

## Verification

- DB clustered startup, restart, forwarding, and three-node replication tests.
- Node-ID persistence across RocksDB reopen.
- Local and leader-side rejection of incompatible protocol/state-machine versions.
- Rejection of wildcard cluster binding without an advertised Raft address.


deven96:
As discussed @jimezesinachi , I took a look and I think like @Iamdavidonuh mentioned... preventing a node from joining an incompatible cluster is maybe a better approach as opposed to creating a persistent identity as that would mean managing some extra statefulness on our end

jimezesinachi:
> As discussed @jimezesinachi , I took a look and I think like @Iamdavidonuh mentioned... preventing a node from joining an incompatible cluster is maybe a better approach as opposed to creating a persistent identity as that would mean managing some extra statefulness on our end

Understood. We can remove the identity stuff and just leave the parts that validate a node when joining a cluster

deven96:
> > As discussed @jimezesinachi , I took a look and I think like @Iamdavidonuh mentioned... preventing a node from joining an incompatible cluster is maybe a better approach as opposed to creating a persistent identity as that would mean managing some extra statefulness on our end
> 
> Understood. We can remove the identity stuff and just leave the parts that validate a node when joining a cluster

Agreed. Let's proceed with that

PR #393, opened 2026-08-24
perf: benchmark metadata string interning strategies

jimezesinachi:
## Summary

- add a Criterion benchmark comparing owned metadata with integer-ID key interning and full string interning
- measure representation build cost, equality-filter performance, response materialization, and live jemalloc allocation
- assert equivalent filter counts and fully materialized metadata across all representations
- add a self-contained runner for reproducible same-ID Criterion comparisons

This PR contains benchmark code only and does not change Ahnlich production storage or behavior.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy -p db --bench metadata_representation -- -D warnings`
- `cargo check -p db --benches`
- `make test` (342 passed, 5 skipped)
- `./benchmarks/run_metadata_representation_ab.sh`

### DIFF
--- a/ahnlich/db/Cargo.toml
+++ b/ahnlich/db/Cargo.toml
@@ -84,3 +84,7 @@ harness = false
 [[bench]]
 name = "persistence"
 harness = false
+
+[[bench]]
+name = "metadata_representation"
+harness = false
--- a/ahnlich/db/benches/metadata_representation.rs
+++ b/ahnlich/db/benches/metadata_representation.rs
@@ -1,0 +1,321 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use ahnlich_types::metadata::{MetadataValue, metadata_value::Value};
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use tikv_jemalloc_ctl::{epoch, stats};
+
+const KEY_CARDINALITY: usize = 4;
+
+#[derive(Clone, Copy)]
+enum Representation {
+    Owned,
+    KeyInterned,
+    Interned,
+}
+
+impl Representation {
+    fn from_env() -> Self {
+        match std::env::var("AHNLICH_METADATA_REPRESENTATION")
+            .as_deref()
+            .unwrap_or("owned")
+        {
+            "owned" => Self::Owned,
+            "key_interned" => Self::KeyInterned,
+            "interned" => Self::Interned,
+            value => {
+                panic!(
+                    "invalid metadata representation '{value}': expected owned, key_interned, or interned"
+                )
+            }
+        }
+    }
+}
+
+type SourceRows = Vec<HashMap<String, MetadataValue>>;
+
+struct OwnedMetadata {
+    rows: SourceRows,
+}
+
+#[derive(Default)]
+struct StringInterner {
+    ids: HashMap<Arc<str>, u32>,
+    values: Vec<Arc<str>>,
+}
+
+impl StringInterner {
+    fn intern(&mut self, value: &str) -> u32 {
+        if let Some(id) = self.ids.get(value) {
+            return *id;
+        }
+
+        let id = u32::try_from(self.values.len()).expect("benchmark interner exhausted u32 IDs");
+        let value: Arc<str> = Arc::from(value);
+        self.values.push(Arc::clone(&value));
+        self.ids.insert(value, id);
+        id
+    }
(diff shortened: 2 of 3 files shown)


jimezesinachi:
## Metadata representation experiment results

I tested three benchmark-local representations:

- **owned:** current shape, `HashMap<String, MetadataValue>`
- **key-interned:** metadata keys become per-store `u32` IDs; `MetadataValue` remains unchanged
- **fully interned:** both keys and `RawString` values become `u32` IDs

### Method

The deterministic fixture contains 1,000, 10,000, or 100,000 rows, with four metadata keys per row and eight repeated raw-string values per key. It uses Ahnlich's generated `MetadataValue` type. Before measuring, the benchmark asserts identical equality-filter counts and identical fully materialized metadata for all representations.

Build measures the additional storage-representation work after request decoding. Criterion excludes fixture cloning from the timed section, so the owned control is nearly free because it only moves the decoded collection.

Environment: commit `a515297c0bf6bdf998bc9873197d5889510e80fe`, Rust 1.97.1, Apple M1 Pro (10 cores), 16 GiB RAM. Criterion uses 10-second measurement windows and 10 samples.

### Live metadata allocation

| Rows | Owned | Key-interned | Fully interned |
|---:|---:|---:|---:|
| 1,000 | 618,744 B | 453,416 B (-26.7%) | 120,920 B (-80.5%) |
| 10,000 | 6,400,040 B | 4,703,168 B (-26.5%) | 1,285,648 B (-79.9%) |
| 100,000 | 63,951,048 B | 47,235,576 B (-26.1%) | 12,795,808 B (-80.0%) |

### Representation build

| Rows | Owned move | Key-interned | Difference | Fully interned | Difference |
|---:|---:|---:|---:|---:|---:|
| 1,000 | 80.0 ns | 380.4 us | +475,605% | 628.7 us | +786,126% |
| 10,000 | 437.1 ns | 3.8 ms | +880,308% | 6.3 ms | +1,440,815% |
| 100,000 | 517.6 ns | 38.4 ms | +7,409,886% | 64.0 ms | +12,356,168% |

Key interning is therefore **475,605-7,409,886% slower** than owned representation build in this benchmark, corresponding to approximately **0.38 us of additional conversion work per row**.

The percentages are unusually large and grow with the fixture because the owned representation performs an O(1) move of already-decoded metadata, while interning performs O(rows x keys) hashing and reconstruction. They describe the isolated representation-conversion regression, not the percentage impact on Ahnlich's complete `Set` path.

### Equality filtering

| Rows | Owned | Key-interned | Fully interned |
|---:|---:|---:|---:|
| 1,000 | 15.9 us | 11.5 us | 9.7 us |
| 10,000 | 159.7 us | 110.3 us | 94.2 us |
| 100,000 | 3.7 ms | 1.26 ms | 969.3 us |

Key interning is approximately **28-66% faster** for this equality-filter workload. Full interning is approximately **39-74% faster**.

### Materialization back to public metadata

| Rows | Owned | Key-interned | Fully interned |
|---:|---:|---:|---:|
| 1,000 | 690.9 us | 751.1 us | 741.1 us |
| 10,000 | 7.1 ms | 7.6 ms | 7.5 ms |
| 100,000 | 71.8 ms | 76.1 ms | 76.5 ms |

Key interning is approximately **6-9% slower** when reconstructing public metadata.

### Interpretation

Full string interning does not look suitable for production: its large memory and filter gains come with broader value conversion, storage-format complexity, and a higher ingestion cost.

Integer-ID key interning is more credible. It keeps every `MetadataValue` unchanged, saves about **26%** of metadata allocation, and improves this equality scan while adding about **0.38 us per ingested row** and 6-9% materialization overhead. That is enough to justify further measurement, but not a production change yet.

The main limitation is that this is an isolated, highly repetitive metadata workload. It does not include protobuf decoding, Raft, vector/index updates, snapshots, or the complete `Set` RPC. If we decide to proceed, the next decision point should be a production-shaped end-to-end `Set` and query benchmark before changing the stored representation.

Reproduce with:

```bash
cargo install critcmp
./benchmarks/run_metadata_representation_ab.sh
```

deven96:
Not sure I'm convinced of this string interning either @jimezesinachi

jimezesinachi:
> Not sure I'm convinced of this string interning either @jimezesinachi



> Not sure I'm convinced of this string interning either @jimezesinachi

Yep, shutting this one down

PR #398, opened 2026-08-27
perf(db): optimize linear search parallelism

jimezesinachi:
## Summary

- avoid cloning every Papaya linear-search candidate by collecting borrowed entries under an owned guard
- select sequential or Rayon search using embedding-dimension bands scaled by active request count
- retain the same guard while materializing the final top-N response
- add an opt-in Criterion workflow that reports changed experimental benchmarks on performance PRs

## Benchmarking

The included `papaya_linear_search` experiment compares the previous cloned staging path against borrowed staging and the dimension-aware policy. Applying the `benchmark-experiments` label runs the experiment workflow and posts the `critcmp` results to the PR.

## Verification

- `cargo test -p db engine::store::tests -- --nocapture`
- `make clippy`
- `cargo bench --manifest-path db/Cargo.toml --bench papaya_linear_search --no-run`
- `cargo fmt --all -- --check`
- `git diff --check`


jimezesinachi:
## SIFT1M A/B benchmark results

I ran the existing baseline benchmark harness against `origin/main` and this PR using the same SIFT1M workload and host configuration.

### System specs

Chip: Apple M1 Pro
Total Number of Cores: 10 (8 performance and 2 efficiency)
Memory: 16 GB

### Configuration

- SIFT1M source, loading 100,000 vectors
- 128-dimensional embeddings
- 2,000 measured requests after 200 warm-up requests
- 3 repeats per configuration; values below are medians
- concurrency: 1, 2, 4, 8, and 16
- Rayon thread pool: 16 threads
- scenarios: ping, unfiltered linear search, filtered linear search (5,000 / 1,000 / 100 candidates), and HNSW
- zero failed requests in every run

### Unfiltered linear search

| Concurrency | Main RPS | PR RPS | RPS change | CPU/request change | p99 change |
|---:|---:|---:|---:|---:|---:|
| 1 | 74.5 | 108.7 | **+45.9%** | **-7.8%** | **-12.7%** |
| 2 | 91.3 | 133.8 | **+46.5%** | **-7.6%** | **-12.4%** |
| 4 | 108.8 | 189.2 | **+73.9%** | **-33.9%** | **-17.8%** |
| 8 | 147.2 | 337.8 | **+129.5%** | **-61.5%** | **-61.4%** |
| 16 | 156.1 | 381.1 | **+144.1%** | **-62.1%** | **-72.7%** |

Filtered linear search follows the same pattern:

- concurrency 1: **+57.1% to +64.8% RPS**
- concurrency 2: **+52.6% to +57.8% RPS**
- concurrency 4: **+89.8% to +95.5% RPS**
- concurrency 8: **+143.9% to +147.2% RPS**
- concurrency 16: **+165.4% to +165.9% RPS**

At concurrency 8 and 16, filtered-search CPU per request fell by roughly **64–65%**, while p99 latency fell by roughly **63–76%**.

### Controls and interpretation

Ping throughput remained within **+0.1% to -2.7%**, so there is no evidence of material client or transport drift favoring the PR.

HNSW was noisy and slower in the PR run (**-15.4% to -27.9% RPS**, with repeat spread as high as 57%). Since this change does not affect HNSW, that points to unfavorable host variance during the PR run. It makes the repeated linear-search gains conservative rather than an artifact of a generally faster second run.

The strongest result is therefore not only higher linear-search throughput, but the matching reduction in server CPU per request and tail latency as concurrency rises.

Full generated summaries:

- [main baseline summary](https://github.com/deven96/ahnlich/blob/6e9b47575e5147a0c16694c1536964aba0dc5ada/benchmarks/results/ab_main_sift100k_final/SUMMARY.md)
- [PR summary](https://github.com/deven96/ahnlich/blob/6e9b47575e5147a0c16694c1536964aba0dc5ada/benchmarks/results/ab_pr_sift100k_final/SUMMARY.md)

These generated summaries are committed temporarily for maintainer review and will be removed before merge.

deven96 (approved):
Amazing improvements!

Issue #397, opened 2026-08-27
Pagination endpoint for Ahnlich DB store


deven96:
Should probably have a list endpoint that allows iterating through a store. We could provide support for predicate conditions as well which effectively gives us an operation like 

```proto
message ListStore {
    string store = 1;
    optional uint64 start_key = 2;  // StoreKeyId perhaps as a token to start from
    uint32 limit = 3;                // Max number of items to return
    optional PredicateCondition condition = 4;  // Filter results by predicate
    optional string schema = 5;      // Schema/namespace
}
```

Given that ordering has no semantic meaning, should we bother to implement ordering. If yes we could still sort by `StoreKeyId` first. 

Checklist follow up to this would be:
-  Integrating endpoint into Ahnlich AI and CLI
- Updating libraries to include new endpoint
- Updating https://ahnlich.dev public docs
- VectorDBZ as #396 stated

osobotu:
I could take a stab at this because it holds back #396 

I propose naming the endpoint `ListStoreEntries` to avoid confusing it with `ListStores`. The pagination implementation will use a cursor based on the deterministic StoreKeyId. The store entries will be ordered internally by StoreKeyId, and each response returns a next_cursor when more entries exist. The cursor should have no semantic meaning to clients, it should be saved and returned to get the next page. We do not need user-configurable ordering for now.

ListStoreEntries should accept a store, schema, limit, optional cursor (or start_key), and optional predicate condition, as you have defined.

Issue #408, opened 2026-08-30
Add a native ClearStore endpoint to Ahnlich DB


osobotu:
VectorDBZ (#396) requires a way to delete all entries in a store without dropping the store itself.

This could currently be done by listing entries and sending their embedding keys back through multiple `DelKey` requests. However, this adds unnecessary network calls and could leave the store partially cleared if one request fails.

I propose adding a native `ClearStore` endpoint that accepts a store name and optional schema. A predicate condition is unnecessary because DelPred already supports conditional deletion.

`ClearStore` should:
- Delete all entries through the internal deletion path.
- Preserve the store name, schema, dimension, creation date, predicate indexes, and non-linear index configuration.
- Return the number of deleted entries.
- Update the persistence state.
- Be replicated as a single mutation.
- Return an appropriate error when the store does not exist.

The existing store should be cleared in place rather than dropped and recreated.

PR #409, opened 2026-08-31
ListStoreEntries and ClearStore DB operations

osobotu:
This PR implements and tests the new DB operations (ListStoreEntries and ClearStore).
Fixes issues #397 and #408.


ahnlich/db/src/engine/store.rs
```diff
@@ -1145,6 +1196,56 @@ impl Store {
...
+                    return None;
+                }
+
+                if matching_ids.as_ref().is_some_and(|ids| !ids.contains(&id)) {
+                    return None;
+                }
+
+                Some((id, embedding_key.clone(), Arc::clone(store_value)))
+            })
+            .collect::<Vec<_>>();
```
deven96, line 1225: Cloning potentially the entire store is undoable.... I need to think

deven96: We could consider maintaining a `BTreeMap` or a crossbeam `SkipSet` where the latter would need a lot more bytes per node (ballpark maybe 32-40 bytes per node i.e 40MB overhead per 1M store entries) but would not require a `RwLock|Mutex`

Insert operations take the hit regardless in either case but it's unavoidable if we want consistent ordering 

In that case, we'd also need to figure out a new implementation i.e a V3 store where we migrate V2 stores into it on snapshot load

ahnlich/db/src/engine/store.rs
```diff
@@ -1145,6 +1196,56 @@ impl Store {
             .collect()
     }
 
+    fn list_entries(
+        &self,
+        cursor: Option<StoreKeyId>,
+        limit: NonZeroUsize,
+        condition: Option<&PredicateCondition>,
+    ) -> Result<StoreEntryPage, ServerError> {
+        let matching_ids = condition
```
deven96, line 1205: As a follow up to the comment below, when implementing predicate filtering with the index, we should use some form of selectivity heuristic to avoid performance issues

An example of such would be we sort the matching IDs from `self.predicate_indices.matches()` directly and iterate over them whenever those matches are small as that would avoid scanning the `SkipSet` entirely.

```rust
if let Some(condition) = condition {
    let matching_ids = self.predicate_indices.matches(condition);
    let total_entries = self.cached_len.load(Ordering::Relaxed) as usize;
    let selectivity = matching_ids.len() as f64 / total_entries.max(1) as f64;

  if selectivity < selectivity_threshold(total_entries, limit) { 
     self.list_entries_sort_matching(cursor, limit, matching_ids)
    } else {
     // otherwise use new sorted index 
     self.list_entries_scan_skipset_with_filter(cursor, limit, &matching_ids)
    }
} else {
    self.list_entries_from_scan_skipset_without_filter(cursor, limit)
}

// Heuristic table to calculate whenever sorting is preferable to scanning
fn selectivity_threshold(total_entries: usize, limit: usize) -> f64 {
    let base_threshold = match total_entries {
        0..=1_000 => 0.10,
        1_001..=10_000 => 0.03,
        10_001..=100_000 => 0.01,
        100_001..=1_000_000 => 0.003,
        _ => 0.001,
    };
    
    // Adjust for limit size: larger limits favor sorting longer
    let limit_factor = ((limit as f64) / 100.0).sqrt();
    (base_threshold * limit_factor).min(0.5)
}
```

deven96 (approved):

PR #412, opened 2026-09-01
perf (db): optimize primary map writes

jimezesinachi:
## Summary

- Reuse one Papaya guard for the complete primary-map write phase of `Set`.
- Process primary-map writes sequentially instead of dispatching each entry through Rayon.
- Replace shared `AtomicUsize` counters with local inserted and updated counters.
- Preserve existing key/value ownership and replacement behavior.
- Leave predicate-index and nonlinear-index processing unchanged, including their existing parallelism policies.

## Why

The previous implementation pinned the primary map once per entry and updated shared sequentially consistent atomic counters from Rayon workers.

Benchmarking showed that Rayon scheduling, repeated guard acquisition, and atomic synchronization cost substantially more than the primary-map writes themselves. A single sequential guard with local counters removes that overhead without changing `Set` behavior.

## Benchmarks

Criterion configuration:

- 128, 768, and 1,536-dimensional embeddings
- three metadata fields per entry
- insert-only, update-only, and mixed workloads
- batch sizes from 100 to 100,000 entries
- 16 Rayon threads
- 20 samples, 1-second warmup, and 3-second measurement
- two complete runs
- fresh pre-sized Papaya map outside each timed iteration

Latency improvement over the current per-entry-pin implementation:

| Entries | Dimensions | Insert | Update | Mixed |
|---:|---:|---:|---:|---:|
| 2k | 128 | 74.7-75.4% | 89.6-89.9% | 89.4-89.6% |
| 2k | 768 | 73.2-73.9% | 89.8-90.0% | 89.0-89.3% |
| 2k | 1,536 | 72.5-73.1% | 87.0-87.8% | 88.9-89.0% |
| 10k | 128 | 63.2-64.2% | 75.8-77.3% | 78.7-80.0% |
| 10k | 768 | 62.2-65.1% | 72.9-73.9% | 79.3-79.5% |
| 10k | 1,536 | 62.3-65.8% | 65.3-67.3% | 73.0-75.0% |
| 25k | 128 | 47.4-50.0% | 51.6-58.1% | 66.1-68.6% |
| 25k | 768 | 49.3-57.2% | 48.1-50.4% | 54.1-61.4% |
| 25k | 1,536 | 52.1-57.3% | 41.3-44.0% | 53.1-54.7% |
| 50k | 128 | 27.8-33.2% | 29.1-34.4% | 40.8-41.0% |
| 50k | 768 | 34.6-38.9% | 26.7-31.6% | 34.2-45.4% |
| 50k | 1,536 | 30.3-38.2% | 21.5-31.5% | 36.7-39.7% |
| 100k | 128 | 1.5-4.7% | 13.1-14.2% | 15.7-17.5% |
| 100k | 768 | 7.2-8.9% | 25.1-25.9% | 24.4-25.6% |
| 100k | 1,536 | 7.6-16.5% | 12.4-13.3% | 17.6-24.7% |

The advantage narrows at larger batch sizes, where allocation destruction and cache pressure become more significant. However, the one-guard implementation remained faster across every tested workload, batch size, dimension, and repeated run.

Memory measurements for a 10,000-entry update workload showed overlapping peak RSS ranges, with no material regression.

## End-To-End Validation

The ghz comparison used SIFT10K and exercised insert-only, update-only, and mixed `Set` workloads across batch sizes from 100 to 2,000 entries, concurrency levels from 1 to 16, and three repeats per configuration. Each measured run processed 10,000 entries.

Throughput change across the 15 workload/batch combinations at each concurrency level:

| Concurrency | Median change | Range |
|---:|---:|---:|
| 1 | +1.3% | +0.4% to +2.7% |
| 2 | +1.7% | +0.9% to +5.1% |
| 4 | +1.8% | +0.1% to +6.0% |
| 8 | +2.1% | -0.2% to +9.6% |
| 16 | +3.1% | -0.7% to +9.7% |

The candidate achieved higher entry throughput in 72 of 75 configurations; one result rounded to flat and two were lower by at most 0.7%. All requests completed successfully. The gain was strongest under higher concurrency, while p95 latency was more variable because the larger batches produced relatively few requests per run.

The full reports are included under:

- [Main ghz summary](https://github.com/deven96/ahnlich/blob/perf/papaya-primary-map-writes/benchmarks/results/set_main_10k_ab/SUMMARY.md)
- [Candidate ghz summary](https://github.com/deven96/ahnlich/blob/perf/papaya-primary-map-writes/benchmarks/results/set_candidate_10k_ab/SUMMARY.md)

## Verification

- DB test suite: 92 passed, 1 skipped
- Focused benchmark Clippy passed with `-D warnings`
- Formatting and `git diff --check` passed
- All benchmark cases completed successfully in both runs


deven96 (approved):
Fairly straightforward change... good to go here

PR #415, opened 2026-09-03
perf(db): optimize predicate index operations

jimezesinachi:
## Summary

Optimizes three predicate-index paths whose work previously scaled with the complete index or input batch rather than the data relevant to the operation.

- Skip predicate staging and metadata filtering when a store has no predicate indexes.
- Resolve indexed `In` predicates through direct value-bucket lookups instead of scanning every bucket.
- Remove deleted IDs only from predicate buckets identified by removed-entry metadata.
- Keep empty predicate buckets unchanged to avoid introducing concurrent cleanup semantics.

### Empty-index ingestion

`Store::add` now checks whether predicate indexes are enabled before allocating the predicate staging vector, cloning each `Arc<StoreValue>`, inspecting metadata, or invoking adaptive metadata filtering.

### Direct indexed `In` lookups

Indexed `In` evaluation now performs one map lookup per requested value and combines the matching ID sets. Runtime therefore scales with requested values and matching IDs instead of total index cardinality.

### Targeted deletion

Primary-map deletion retains metadata for entries actually removed. Predicate cleanup uses that metadata to locate the exact value buckets containing each ID. Missing and duplicate requested IDs do not produce additional cleanup work. Empty buckets are deliberately retained.

### Empty-index ingestion

The candidate consistently reduced the skipped predicate phase to approximately `4.92 ns`.

| Entries | Metadata fields | Sequential control | Adaptive control |
|---:|---:|---:|---:|
| 100 | 0 | 38.135 µs | 38.285 µs |
| 1,000 | 4 | 197.05 µs | 195.46 µs |
| 10,000 | 16 | 3.7976 ms | 357.47 ms |
| 100,000 | 0 | 1.5879 ms | 944.17 µs |
| 100,000 | 16 | 41.576 ms | 3.5502 s |

These are phase-level savings. Complete `Set` performance still includes validation, key hashing, embedding preparation, and primary-map insertion.

### Direct `In` lookup

Representative results at `100,000` distinct indexed values:

| Request | Control | Candidate | Reduction |
|---|---:|---:|---:|
| 1 hit, 1 ID/value | 1.2067 ms | 91.332 ns | 99.992% |
| 4 hits, 1 ID/value | 2.9590 ms | 415.68 ns | 99.986% |
| 32 hits, 1 ID/value | 12.520 ms | 3.8421 µs | 99.969% |
| 32 misses, 1 ID/value | 6.7117 ms | 1.1719 µs | 99.983% |
| 32 mixed, 1 ID/value | 9.5686 ms | 2.3114 µs | 99.976% |
| 32 hits, 32 IDs/value | 18.389 ms | 57.236 µs | 99.689% |

Direct lookup avoids adding deduplication overhead to ordinary requests. Duplicate-heavy, low-cardinality inputs can be less favorable, but tested deduplicating and hybrid strategies degraded the more common unique-value cases.

### Targeted deletion

| Scenario | Control | Candidate | Reduction |
|---|---:|---:|---:|
| Low-cardinality sparse | 13.622 µs | 828.10 ns | 93.92% |
| High-cardinality sparse | 3.1733 s | 536.92 ns | 99.99998% |
| High-cardinality batch | 3.2365 s | 98.167 µs | 99.9970% |
| Low-cardinality dense | 148.91 ms | 8.4389 ms | 94.33% |
| Many predicate indexes | 327.68 ms | 129.19 µs | 99.9606% |
| Wide metadata | 154.06 ms | 84.621 µs | 99.9451% |

## Verification

- `cargo test -p db --lib`: 93 passed, 1 ignored
- DB production Clippy passed with warnings denied
- All three experimental benchmark targets passed Clippy
- Benchmark fixtures assert equivalent results between control and candidate implementations

### DIFF
--- a/ahnlich/db/src/engine/predicate.rs
+++ b/ahnlich/db/src/engine/predicate.rs
@@ -163,11 +163,28 @@ impl PredicateIndices {
     }
 
     /// Removes a store key id when it's corresponding entry in the store is removed
+    #[cfg(feature = "bench-experiments")]
     #[tracing::instrument(skip(self))]
     pub(super) fn remove_store_keys(&self, remove_keys: &[StoreKeyId]) {
         let pinned = self.inner.pin();
         for (_, values) in pinned.iter() {
             values.remove_store_keys(remove_keys);
+        }
+    }
+
+    /// Removes each store key id from the predicate buckets described by its stored metadata.
+    #[tracing::instrument(skip_all)]
+    pub(super) fn remove_store_entries<'a>(
+        &self,
+        removed: impl IntoIterator<Item = (StoreKeyId, &'a StoreValue)>,
+    ) {
+        let indices = self.inner.pin();
+        for (store_key_id, store_value) in removed {
+            for (metadata_key, metadata_value) in &store_value.value {
+                if let Some(index) = indices.get(metadata_key) {
+                    index.remove_store_key(metadata_value, &store_key_id);
+                }
+            }
         }
     }
 
@@ -409,6 +426,7 @@ impl PredicateIndex {
     }
 
     /// Removes a store key id when it's corresponding entry in the store is removed
+    #[cfg(any(test, feature = "bench-experiments"))]
     #[tracing::instrument(skip(self))]
     fn remove_store_keys(&self, remove_keys: &[StoreKeyId]) {
         let inner = self.0.pin();
@@ -417,6 +435,13 @@ impl PredicateIndex {
             for k in remove_keys {
                 values_pinned.remove(k);
             }
+        }
+    }
+
+    fn remove_store_key(&self, metadata_value: &MetadataValue, store_key_id: &StoreKeyId) {
+        let buckets = self.0.pin();
+        if let Some(store_key_ids) = buckets.get(metadata_value) {
+            store_key_ids.pin().remove(store_key_id);
         }
     }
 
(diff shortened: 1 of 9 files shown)


ahnlich/db/src/engine/predicate.rs
```diff
@@ -318,6 +335,11 @@ impl PredicateIndices {
         }
     }
 
+    #[inline]
+    pub(super) fn is_empty(&self) -> bool {
+        self.allowed_predicates.is_empty()
```
deven96, line 340: A niche case could be that self.allowed_predicates is not empty but self.inner is empty.... possible if you've created allowed_predicates but not used any metadata containing fields in there

ahnlich/db/src/engine/predicate.rs
```diff
@@ -163,6 +163,7 @@ impl PredicateIndices {
     }
 
     /// Removes a store key id when it's corresponding entry in the store is removed
+    #[cfg(feature = "bench-experiments")]
```
deven96, line 166: I like the feature setup to prove the hypothesis and I'm guessing we are getting rid of them before merge of this PR once confirmed

deven96:
> Duplicate-heavy, low-cardinality inputs can be less favorable

Out of curiousity, how much less favourable?

ahnlich/db/src/engine/store.rs
```diff
@@ -1026,17 +1026,33 @@ impl Store {
...
-            .flat_map(|k| pinned.remove(k))
-            .map(|(k, _)| k.clone())
+            .filter_map(|store_key_id| {
+                pinned
+                    .remove(&store_key_id)
+                    .map(|(embedding_key, store_value)| {
+                        (store_key_id, embedding_key.clone(), Arc::clone(store_value))
+                    })
+            })
             .collect::<Vec<_>>();
```
deven96, line 1038: Could we collect once into two Vecs with `unzip()` i.e `(Vec<(_, _)>, Vec<_>)` here rather than allocating this `Vec<(_, _, _)>` and then down the line collecting again to be able to pass it down into `self.non_linear_indices.delete`

deven96 (approved):

PR #419, opened 2026-09-07
Audio Chunking + poorman's shazam as an example

deven96:


Ayobami-00:
I went through this and I have some comments/suggestions summarized here.
1. Avoid duplicating original inputs for OneToMany responses
   In [single_stage.rs](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/ahnlich/ai/src/engine/ai/providers/ort/single_stage.rs#L487-L495), audio is returned as OneToMany. The generic storage path in [store.rs](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/ahnlich/ai/src/engine/store.rs#L397-L420) clones the complete StoreValue for every generated embedding. With store_original=true, this duplicates the complete image or audio for every face or chunk. A ten-minute audio file could therefore be copied up to 67 times.
   
   Maybe we could store the original input once under a stable source ID, then attach only the source ID, child index, and model metadata to each embedding. For now maybe we reject store_original=true for chunked audio.
   
2. Enforce a separate duration limit for audio queries
   The recording command in [main.go](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/examples/go/audio-search/main.go#L66-L71) correctly limits recordings to ten seconds. However, [search --file](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/examples/go/audio-search/main.go#L195-L210) submits the complete supplied file. Files longer than ten seconds generate multiple embeddings, while the current query path in [store.rs](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/ahnlich/ai/src/engine/store.rs#L568-L575) requires exactly one. This would result in an error. We could allow ten-minute chunking for indexing, but reject query audio longer than ten seconds before inference. The example should validate or trim search --file inputs, and the documentation should distinguish indexing limits from query limits.
    
3. Batch audio chunks before running ONNX inference
   In [single_stage.rs](https://github.com/deven96/ahnlich/blob/4bd7dbc52cce4cf60e1d866a2d349964599df453/ahnlich/ai/src/engine/ai/providers/ort/single_stage.rs#L438-L443), batch_inference_audio is called separately for every chunk even though CLAP is configured with a batch size of eight. A ten-minute file therefore performs 67 sequential ONNX calls instead of approximately nine. We can give every chunk an explicit source index, flatten the chunks, and process them using chunks(self.model_batch_size). Concatenate up to eight feature tensors along the batch axis, run ONNX once, and regroup the returned embeddings by source index into the appropriate OneToMany responses.

PR #421, opened 2026-09-10
perf(similarity): reuse HNSW distance calculations during search

jimezesinachi:
## What changed

HNSW search currently calculates the distance between the query and the same node more than once:

- `search_layer` calculates it when deciding whether to explore a neighbor.
- `NearestFirst::push` calculates it again when adding that neighbor to the candidate queue.
- Scores are discarded between HNSW layers and recalculated when building the next search queue.

This PR adds a search implementation that carries `OrderedNode` values through these stages, allowing the queues and subsequent layers to reuse the scores already calculated.

The Criterion benchmark compares:

- `control`: existing `knn_search`
- `candidate`: distance-reusing search

The benchmark covers 128d, 768d, and 1536d vectors; Euclidean distance, cosine similarity, and dot-product similarity; and `ef` values of 50, 100, and 200.

The ghz harness also accepts `SCENARIOS` and `SERVER_BIN` overrides, allowing selected workloads to run against independently built control and candidate binaries. Each run records the selected binary and its SHA-256 hash.

## Results

The final local Criterion run showed:

| Dimensions | Candidate throughput |
|---|---:|
| 128d | 1.00–1.17× |
| 768d | 1.42–1.47× |
| 1536d | 1.58–1.69× |

The benefit increases with vector dimensionality, where repeated distance calculations account for more of the search cost. No regression was observed across the 27 benchmark cases; 128d Euclidean results were effectively neutral.

## Validation

- Final local Criterion control-versus-candidate run completed across all 27 cases.
- `cargo fmt --all --manifest-path ahnlich/Cargo.toml`
- `git diff --check`


deven96 (approved):
I agree with the sentiment of the change. LGTM

PR #424, opened 2026-09-16
perf(db): share collectors within predicate indexes

jimezesinachi:
## Summary

Reduce predicate-index memory usage by sharing one Papaya collector between the outer map and all value buckets belonging to the same predicate index.

Previously, every predicate-value bucket owned an independent collector. High-cardinality indexes could therefore create tens or hundreds of thousands of collectors, with the collector overhead dominating the memory used by the indexed values themselves.

The new per-index boundary keeps separate predicate indexes isolated while removing the collector-per-bucket cost.

## Implementation

- Introduce `SharedPredicateIndex`, which owns:
  - the outer `MetadataValue -> HashSet<StoreKeyId>` map;
  - one shared Seize collector;
  - inner bucket sets constructed from that collector.
- Preserve the existing predicate lookup, insertion, and deletion behavior.
- Keep the previous independent-collector implementation behind `bench-experiments` as the benchmark control.
- Add fallible constructors for Papaya maps and sets using a shared collector.
- Update DB size expectations for Papaya 0.2.5’s smaller map representation.

## Persistence Compatibility

The collector is runtime state and must not become part of persisted snapshots.

`SharedPredicateIndex` therefore serializes exactly its inner map, preserving the previous serialized shape. During deserialization it:

1. creates a new collector for the predicate index;
2. reconstructs the outer map with that collector;
3. reconstructs every value bucket with the same collector;
4. restores the serialized store-key memberships.

Coverage verifies this behavior through:

- direct predicate-index serialization round trips;
- legacy JSON persistence migration;
- standalone `utils::Persistence` recovery;
- in-memory Raft snapshot recovery;
- durable RocksDB-backed Raft snapshot recovery;
- indexed reads and writes after restoration.

## Papaya Safety Fix

Benchmarking shared collectors exposed an unsafe Papaya drop path: dropping one map called `Collector::reclaim_all()` even when another map sharing that collector still had an active guard.

The issue and reproducer are documented in:

- https://github.com/ibraheemdev/papaya/issues/102

The proposed upstream fix is:

- https://github.com/ibraheemdev/papaya/pull/103

Until that fix is merged and released, this PR pins Papaya to the exact reviewed commit:

```text
1d914b94d846da6185380b814d09758624fd9f7e
```

## Performance

Criterion compares the old independent-collector representation against the new per-index representation using the production insertion implementations.

Representative sequential results:

| Workload | Control | Candidate | Change |
|---|---:|---:|---:|
| Construct empty index | 478 ns | 842 ns | 76.2% slower |
| First unique 10k | 14.239 ms | 10.468 ms | 26.5% faster |
| First unique 100k | 243.30 ms | 128.11 ms | 47.3% faster |
| Low-cardinality 10k | 1.482 ms | 1.490 ms | effectively neutral |

The fixed construction cost increases by approximately 364 ns per predicate index. High-cardinality sequential ingestion improves substantially because it no longer creates an independent collector for every bucket.

Forced parallel high-cardinality insertion remains worse for the candidate:

| Workload | Control | Candidate | Change |
|---|---:|---:|---:|
| Unique 10k parallel | 59.209 ms | 62.351 ms | 5.3% slower |
| Unique 100k parallel | 257.44 ms | 478.96 ms | 86.0% slower |

The benchmark forces these paths for analysis. Production’s current 150k parallel threshold processes the measured 10k and 100k workloads sequentially.

## Memory

Memory measurements run each control/candidate case in fresh processes and report the median of three runs.

### Populated stores

| Scenario | Collectors C / P | Allocated MiB C / P | Saving |
|---|---:|---:|---:|
| Empty | 4 / 4 | 0.142 / 0.142 | 0.0% |
| 1 index, low-cardinality 10k | 15 / 5 | 1.579 / 1.420 | 10.1% |
| 1 index, unique 10k | 10,005 / 5 | 189.789 / 23.853 | 87.4% |
| 4 indexes, unique 10k | 40,008 / 8 | 755.925 / 91.992 | 87.8% |
| 1 index, unique 100k | 100,005 / 5 | 1,891.478 / 231.329 | 87.8% |

### After deleting store keys

| Scenario | Allocated MiB C / P | Saving |
|---|---:|---:|
| 1 index, low-cardinality 10k | 1.256 / 1.095 | 12.8% |
| 1 index, unique 10k | 199.366 / 23.412 | 88.3% |
| 4 indexes, unique 10k | 795.905 / 90.912 | 88.6% |
| 1 index, unique 100k | 1,989.454 / 226.739 | 88.6% |

The report also exposes immediate post-drop retention in the candidate’s deferred Papaya table retirement. For example, the unique 100k case retains approximately 225.6 MiB immediately after dropping the store. This is visible in the report for follow-up alongside the upstream reclamation fix.

## Benchmark Reporting

Extend the opt-in `benchmark-experiments` workflow to support two report types:

- Criterion control/candidate timing comparisons;
- isolated control/candidate memory reports.

Changed memory targets under `benches/memory/` generate Markdown that is appended to the existing update-in-place PR benchmark comment.

## Validation

- Targeted Clippy with `-D warnings`
- Criterion and memory benchmark targets compile
- Full isolated memory matrix completes
- Serialization and restoration coverage added for standalone and clustered persistence


deven96 (approved):
LGTM

Approved but pending removal of the bootstrap code in the core engine

PR #431, opened 2026-09-22
perf(db): use predicate-index for filtered linear search

jimezesinachi:
## Problem

Filtered linear `GETSIMN` currently walks the primary store and evaluates the predicate against each entry. It avoids distance calculations for entries that fail the filter, but still visits every entry when an indexed predicate matches only a small subset. An indexed equality matching 100 entries in a 10,000-entry store, for example, still requires a 10,000-entry scan.

This PR adds a candidate that starts from predicate-index IDs when it can produce a bounded candidate superset. It looks up those entries in the primary store, rechecks the full predicate, and scores the matches. The existing `get_sim_in_store()` remains intact as the current control. Regular builds continue to call it; builds with `bench-experiments` route `GETSIMN` through the candidate for end-to-end testing.

## Candidate behavior

The candidate plans indexed `Equals` and `In` predicates and recursive combinations:

- `And` can use an indexed branch even if the other branch is unindexed: the indexed IDs still contain every possible match, and the full predicate is checked before scoring.
- `Or` uses index IDs only when both branches can supply complete candidate sets. A partial `Or` falls back to the existing scan.
- `NotEquals`, `NotIn`, absent indexes, and plans exceeding the candidate budget fall back to scanning.

The experimental budget is 10% of the store, capped at 10,000 IDs. This bounds candidate collection and leaves broad filters on the existing path.

Separately, this PR corrects `PredicateIndices::matches()` for indexed `NotEquals` and `NotIn`: entries without the metadata field now match those negative predicates, consistent with inline predicate evaluation. This correctness change is not feature gated.

## Benchmarks

Two independent Criterion targets allow CI to publish separate result comments:

1. **Current vs candidate:** the existing full-scan implementation against the bounded candidate.
2. **Historical vs candidate:** a reconstruction of the earlier indexed `get_matches_with_ids()` path against the bounded candidate. This compares implementations; it is not a benchmark of an old release binary.

The SIFT `GETSIMN` fixture now supports indexed equality, `In`, recursive `And`/`Or`, partial expressions, and empty-bucket cases. The ghz A/B used separate control and candidate binaries with SIFT10k, 10,000 stored vectors, Euclidean distance, `closest_n=10`, 10,000 requests per scenario, three repeats, and concurrency 1, 10, and 50. This table shows **median RPS, candidate vs control**, at concurrency 10 and 50:

| Filter | Matches | Concurrency 10 | Concurrency 50 |
|---|---:|---:|---:|
| Indexed `Equals` | 1% | 4,623 → 5,460 (**+18%**) | 5,310 → 6,188 (**+17%**) |
| Indexed `Equals` | 10% | 4,679 → 5,657 (**+21%**) | 5,485 → 6,420 (**+17%**) |
| Indexed `In` | 1% | 4,525 → 5,433 (**+20%**) | 5,252 → 6,148 (**+17%**) |
| Indexed `In` | 10% | 4,557 → 5,555 (**+22%**) | 5,324 → 6,269 (**+18%**) |
| Recursive indexed `And` | 1% | 4,486 → 5,404 (**+20%**) | 5,111 → 5,900 (**+15%**) |
| Partial `And` | 10% | 4,378 → 5,472 (**+25%**) | 5,143 → 6,160 (**+20%**) |
| Recursive indexed `Or` | 1% | 3,648 → 5,414 (**+48%**) | 4,208 → 5,900 (**+40%**) |
| Recursive indexed `Or` | 10% | 3,674 → 5,222 (**+42%**) | 4,262 → 5,829 (**+37%**) |
| Empty indexed bucket | 0% | 13,868 → 24,548 (**+77%**) | 17,270 → 34,682 (**+101%**) |

The 50%-match and partial-`Or` cases take the scan path and were generally within about 2% of control. Unfiltered linear search measured **1.6% lower RPS at concurrency 10** and **2.5% lower at concurrency 50**, so fallback overhead remains part of the production-migration decision. One indexed-`Or` case was also slower than the reconstructed historical path in Criterion; the end-to-end comparison above is against current main.

## Validation and scope

The preflight answer comparison matched across **19 scenarios and 1,900 queries**. The ghz runs recorded no failed requests. Both Criterion targets completed with their answer assertions passing, as did the focused integration tests and DB library tests.

This PR establishes the candidate and its measurements. It does not switch normal-build linear `GETSIMN` to the candidate; production migration can follow review of these results.


deven96 (approved):
LGTM!

Really cool gains on filtered queries.... we can clean things up and get this merged

